PR #5, opened 2019-10-21
adding initial dockerfile

Rabscuttle:


docker/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM centos
+
+RUN mkdir /opt/tomcat/
+
+WORKDIR /opt/tomcat
+RUN curl -O https://archive.apache.org/dist/tomcat/tomcat-7/v7.0.96/bin/apache-tomcat-7.0.96.tar.gz
```
Rabscuttle: I dont like leaving this external dependency in the build - it will cause problems if apache is down at the time we need to build...We could keep it locally on an s3 bucket, or switch to a tomcat specific docker image.

PR #6, opened 2019-10-22
Feature/docker

Rabscuttle:


lib/context.xml
```diff
@@ -0,0 +1,39 @@
...
+          initialSize="10"
+          removeAbandonedTimeout="60"
+          removeAbandoned="true"
+          logAbandoned="true"
+          minEvictableIdleTimeMillis="30000"
+          jmxEnabled="true"
+          jdbcInterceptors="org.apache.tomcat.jdbc.pool.interceptor.ConnectionState;
+            org.apache.tomcat.jdbc.pool.interceptor.StatementFinalizer"
+          username="root"
+          password="z0g*DR56"
```
nicole-hunter: I'm not super familiar with docker so maybe this is supposed to be like this, but do we want someones actual root password in here?

nicole-hunter: Just kidding, this is what I get for looking at this commit by commit. I see it's removed in the next commit. Disregard!

nicole-hunter:
Looks good to me.

PR #7, opened 2019-10-25
Feature/feature flag

nicole-hunter:


src/main/java/liquibase/featureFlagDataBaseChanges.xml
```diff
@@ -0,0 +1,20 @@
...
+            `FLAG_NAME` VARCHAR(100) NOT NULL,
+            `FLAG_ON` BOOLEAN NOT NULL,
+            PRIMARY KEY (`ID`)
+            ) ENGINE=InnoDB DEFAULT CHARSET=latin1;
+
+            INSERT INTO `FEATURE_FLAGS` (FLAG_NAME, FLAG_ON) VALUES ('MAT_ON_FHIR', 1);
+        </sql>
+    </changeSet>
+
+</databaseChangeLog>
```
adongare: @AndrewBird81 @nicole-hunter: just trying to understand the feature flag functionality. my understanding is when this flag is on, `MAT` will be `MAT on FHIR` and using this flag(like `isFlagOn` or similar) we can enable/disable FHIR specific functionality. Considering this is going to be temporary, can we not achieve this just by putting one property in MAT.properties file(say mat.fhir.mode = on/true) and access it wherever needed?

nicole-hunter: We definitely could do that (and it would have been much easier)! However I think the point of the story was also to set up the frame work for future feature flags as well as doing the mat on FHIR feature flag

adongare: gotcha! Thanks

greenemcg: If we have many potential new feature flags, it is still easier and cleaner to use a property file.

I would create a java class (spring component) of these properties and have the GWT service return this.

Then we do add new properties we just add them to the class we created,  and we would not be limited to just booleans.

Plus these flags control the application and are best stored in the application to be less confusing, and no requirements exist to change these while application is running.

ferret1964: So as I see it using a database and a service give us a broad range of flexibility. As I understand it the notion is that application features might we be tied to specific users sessions in the future, hence using a service tied to the current context makes sense.

nicole-hunter: I see both sides to this argument. The acceptance criteria of my story stated to use liquibase to make these changes which is why I went the DB route. I can switch it up to do the property file if that is what we would prefer. @AndrewBird81 I think we may need you to make the final call here.

PR #12, opened 2019-10-30
Add Model Type to new/draft/cloned Measures

adongare:
Added Model Type(FHIR/QDM) in create measure and create composite  measure view
1. Create Measure view

MeasureModel1

2. Create Composite Measure View

MeasureModel2

In progress: Updating test cases...

### DIFF
--- a/src/main/java/liquibase/addMeasureModelColumn.xml
+++ b/src/main/java/liquibase/addMeasureModelColumn.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+
+    <changeSet author="mat_dev_user" id="1" context="prod">
+        <sql>
+            ALTER TABLE MEASURE ADD COLUMN MEASURE_MODEL VARCHAR(10) NULL;
+            <!-- all the mat v5.x measures are supposed to be QDM type. Old measures(mat < v5.x) will have null value -->
+            UPDATE MEASURE SET MEASURE_MODEL = "QDM" WHERE LOWER(RELEASE_VERSION) LIKE "v5%";
+        </sql>
+    </changeSet>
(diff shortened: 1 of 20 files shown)


src/main/java/liquibase/addMeasureModelColumn.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+
+    <changeSet author="mat_dev_user" id="1" context="prod">
+        <sql>
+            ALTER TABLE MEASURE ADD COLUMN MEASURE_MODEL VARCHAR(10) NULL;
```
AndrewBird81: Should this be required? Does it need a default value, for all the legacy data?

adongare: yeah, I'm thinking this should be default to QDM. But @nicole-hunter just mentioned all existing measures may not be QDM type. so, I'm bit confused

nicole-hunter: Only like 1500 measures have the QDM version filled out which makes me think the older ones aren't qdm.

src/main/java/mat/client/measure/AbstractNewMeasureView.java
```diff
@@ -2,6 +2,9 @@
 
 import java.util.List;
 
+import com.google.gwt.user.client.Window;
```
nicole-hunter: I think this import was just here for testing and we shouldn't need it any more.

adongare: that's right, will remove all unused imports. Thanks!

src/main/java/liquibase/addMeasureModelColumn.xml
```diff
@@ -0,0 +1,14 @@
...
+            <!-- all the mat v5.x measures are supposed to be QDM type. Old measures(mat < v5.x) will have null value -->
+            UPDATE MEASURE SET MEASURE_MODEL = "QDM" WHERE LOWER(RELEASE_VERSION) LIKE "v5%";
```
adongare: This was decided as per latest discussion between Nicole, Kathy and Cynthia

PR #13, opened 2019-11-01
adding cql library type to new cql library/draft screens

nicole-hunter:


src/main/java/liquibase/addLibraryModelColumn.xml
```diff
@@ -0,0 +1,14 @@
...
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+
+    <changeSet author="mat_dev_user" id="1" context="prod">
+        <sql>
+            ALTER TABLE CQL_LIBRARY ADD COLUMN LIBRARY_MODEL VARCHAR(10) NULL;
+            <!-- all the mat v5.x measures are supposed to be QDM type. Old measures(mat < v5.x) will have null value -->
+            UPDATE CQL_LIBRARY SET LIBRARY_MODEL = "QDM";
```
adongare: I think message in 10 needs to be updated with respect to library. All the preqdm libs will be of QDM type?

nicole-hunter: good catch on the message. I will update that! And since all cql libraries came after cql they will all say QDM. Measures existed before cql which made the measure side more complex.

ddecouteau:
Looks Good

PR #14, opened 2019-11-05
Add edit status of Measure & CQL Library to Recent Activity &…

adongare:
1. Display the edit status of Measure- both Recent activity and All Measures grid

measurelibview

2. Display the edit status of CQL Library- both Recent activity and My CQL Libraries grid

CQLibView

### DIFF
--- a/src/main/java/mat/client/shared/CQLLibraryResultTable.java
+++ b/src/main/java/mat/client/shared/CQLLibraryResultTable.java
@@ -257,16 +257,19 @@ public class CQLLibraryResultTable {
 	private SafeHtml getCQLLibraryNameColumnToolTip(CQLLibraryDataSetObject object) {
 		SafeHtmlBuilder sb = new SafeHtmlBuilder();
 		String cssClass = "customCascadeButton";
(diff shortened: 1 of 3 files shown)


src/main/java/mat/client/shared/CQLLibraryResultTable.java
```diff
@@ -275,6 +278,30 @@ private SafeHtml getCQLLibraryNameColumnToolTip(CQLLibraryDataSetObject object)
 		}
 		return sb.toSafeHtml();
 	}
+
```
adongare: General question @nicole-hunter @AndrewBird81 : Is there any specific reason to have duplicate code/methods? I've seen almost similar code in measure screens and library screens. if you see two new methods that I added for this PR are almost same.  we can have some utility class and move common code there

nicole-hunter: I think when developing it had been decided they were different enough that we would leave them separate. Other than that we don't really have a big reason. However we need to do a cost/benefit of taking the time now to rework it as this code could potentially all be gone in the near future.

adongare: got it, thanks!

PR #16, opened 2019-11-18
Measure Library - Recent Activity Grid - Add Measure Type to Grid Row

RohitKandimalla:
Measure Library - Recent Activity Grid - Add Measure Type to Grid Row

image

Measure Library - My Measures Grid - Add Measure Type to Grid Row

image


RohitKandimalla:
> This is looking really great! I think that we still need to address the "Pre-CQL" measure type for measures created prior to CQL though. This would be any measure created prior to MAT version 5!

Updated with suggested changes.

RohitKandimalla:
> This is looking really great! I think that we still need to address the "Pre-CQL" measure type for measures created prior to CQL though. This would be any measure created prior to MAT version 5!

Updated with suggested changes. Please Review

PR #21, opened 2019-12-04
Hotfix for MAT 5.8.3

nicole-hunter:
Adding a function to not strip white spaces out while reverse engineering the cql file. This will fix the tuple bug the MAT is experiencing

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>CMAT</groupId>
     <artifactId>MeasureAuthoringTool</artifactId>
-    <version>6.0</version>
+    <version>5.8</version>
     <packaging>war</packaging>
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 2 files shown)


src/main/java/mat/server/cqlparser/ReverseEngineerListener.java
```diff
@@ -213,6 +214,12 @@ public void enterFunctionDefinition(FunctionDefinitionContext ctx) {
 		cqlModel.getCqlFunctions().add(function);
 	}
 
+	private String getFullText(ParserRuleContext context) {
```
AndrewBird81: Should this be static?

pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>CMAT</groupId>
     <artifactId>MeasureAuthoringTool</artifactId>
-    <version>5.8</version>
+    <version>6.0</version>
```
nicole-hunter: This needs to be switched back to 5.8, waiting to see if other parts of this file need updated/changed too.

PR #23, opened 2019-12-10
Feature/mat 62 67

RohitKandimalla:
An incomplete User story, Acceptance criteria to uncheck Checkbox on click of a row element hasn’t been implemented.

### DIFF
--- a/src/main/java/mat/client/shared/CQLLibraryResultTable.java
+++ b/src/main/java/mat/client/shared/CQLLibraryResultTable.java
@@ -32,10 +32,8 @@ import mat.client.util.CellTableUtility;
 import mat.model.cql.CQLLibraryDataSetObject;
 import mat.shared.ClickableSafeHtmlCell;
 
-import java.util.Set;
-
 public class CQLLibraryResultTable {
-	
+
 	private Observer observer;
 
 	public CellTable<CQLLibraryDataSetObject> addColumnToTable(String Label, CellTable<CQLLibraryDataSetObject> table, HasSelectionHandlers<CQLLibraryDataSetObject> fireEvent) {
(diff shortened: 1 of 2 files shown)


src/main/java/mat/client/shared/CQLLibraryResultTable.java
```diff
@@ -1,18 +1,23 @@
 package mat.client.shared;
 
+import com.google.gwt.cell.client.*;
+import com.google.gwt.event.dom.client.ClickEvent;
+import com.google.gwt.event.dom.client.ClickHandler;
+import com.google.gwt.event.dom.client.DoubleClickEvent;
+import com.google.gwt.event.dom.client.DoubleClickHandler;
+import com.google.gwt.event.logical.shared.SelectionEvent;
+import com.google.gwt.user.client.Window;
```
nicole-hunter: This is an unused import i'm pretty sure.

src/main/java/mat/client/shared/CQLLibraryResultTable.java
```diff
@@ -47,12 +66,15 @@ public SafeHtml getValue(CQLLibraryDataSetObject object) {
 				return getCQLLibraryNameColumnToolTip(object);
 			}
 		};
-		cqlLibraryName.setFieldUpdater(new FieldUpdater<CQLLibraryDataSetObject, SafeHtml>() {
+//		Double Click event on Grid row to navigate to Composer Tab.
+		table.addDomHandler(new DoubleClickHandler() {
```
nicole-hunter: I think we need to wrap the double click handler in the feature flag as well

src/main/java/mat/model/cql/CQLLibraryDataSetObject.java
```diff
@@ -39,7 +39,18 @@
 	private boolean isVersionable;
 	private boolean isDeletable;
 	private String libraryModelType;
-	
+	private boolean isCheckboxSelected = false;
```
nicole-hunter: I know we talked about this approach but it doesn't look like you are utilizing it currently. If you think you'll need it for the final solution, lets leave it! Otherwise we should remove it now so we don't accidentally keep it in.

src/main/java/mat/client/shared/CQLLibraryResultTable.java
RohitKandimalla: you are right, I will update this asap.

src/main/java/mat/model/cql/CQLLibraryDataSetObject.java
RohitKandimalla: I will make a note to remove it in future commit, If not being used.

PR #28, opened 2019-12-16
MAT 62 - 67 Revised with Single Click event

RohitKandimalla:
Differentiated Single Click and Double Click using SystemTime with 500ms as ideal threshold for consecutive clicks. 

'onBrowserEvent' is performed on a column instead of a row. So it is added for LibraryName Column, Version Column and Model Type Column. Hasn't implemented on other columns since they will be removed in MAT 63 & 68.

image

Note: removed the unwanted lines of code and previous implementations/ideas.

### DIFF
--- a/src/main/java/mat/client/shared/CQLLibraryResultTable.java
+++ b/src/main/java/mat/client/shared/CQLLibraryResultTable.java
@@ -1,22 +1,12 @@
 package mat.client.shared;
 
 import com.google.gwt.cell.client.CheckboxCell;
-import com.google.gwt.event.dom.client.DoubleClickEvent;
-import com.google.gwt.event.dom.client.DoubleClickHandler;
 import com.google.gwt.cell.client.*;
 import com.google.gwt.event.logical.shared.SelectionEvent;
-import com.google.gwt.view.client.SingleSelectionModel;
 import mat.shared.model.util.MeasureDetailsUtil;
-import org.gwtbootstrap3.client.ui.constants.ButtonType;
-import org.gwtbootstrap3.client.ui.gwt.ButtonCell;
-
-import com.google.gwt.cell.client.Cell.Context;
-import com.google.gwt.dom.client.BrowserEvents;
 import com.google.gwt.dom.client.Element;
 import com.google.gwt.dom.client.NativeEvent;
-import com.google.gwt.dom.client.TableCaptionElement;
 import com.google.gwt.event.logical.shared.HasSelectionHandlers;
-import com.google.gwt.event.logical.shared.SelectionEvent;
 import com.google.gwt.safehtml.shared.SafeHtml;
 import com.google.gwt.safehtml.shared.SafeHtmlBuilder;
 import com.google.gwt.safehtml.shared.SafeHtmlUtils;
@@ -28,7 +18,6 @@ import mat.client.cql.CQLLibrarySearchView.Observer;
 import mat.client.util.CellTableUtility;
 import mat.model.cql.CQLLibraryDataSetObject;
 import mat.shared.ClickableSafeHtmlCell;
-import mat.shared.model.util.MeasureDetailsUtil;
 
 public class CQLLibraryResultTable {
 
--- a/src/main/java/mat/client/shared/CQLLibraryResultTable.java
+++ b/src/main/java/mat/client/shared/CQLLibraryResultTable.java
@@ -1,10 +1,8 @@
 package mat.client.shared;
 
 import com.google.gwt.cell.client.*;
-import com.google.gwt.event.dom.client.DoubleClickEvent;
-import com.google.gwt.event.dom.client.DoubleClickHandler;
 import com.google.gwt.event.logical.shared.SelectionEvent;
-import com.google.gwt.view.client.MultiSelectionModel;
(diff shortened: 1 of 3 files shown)


src/main/java/mat/client/shared/CQLLibraryResultTable.java
```diff
@@ -62,41 +64,75 @@ public Boolean getValue(CQLLibraryDataSetObject object) {
...
-					for (CQLLibraryDataSetObject object : selectionModel.getSelectedSet()) {
-						SelectionEvent.fire(fireEvent, object);
-					}
+			/*
+				Single Click to select row and enable checkbox
+				Double Click to navigate to CQL Composer tab
+			*/
+			@Override
+			public void onBrowserEvent(Cell.Context context, Element elem, CQLLibraryDataSetObject object, NativeEvent event) {
+				if (lastClick < System.currentTimeMillis() - 500)
```
serhii-ilin: Doesn't comply with DRY https://en.wikipedia.org/wiki/Don%27t_repeat_yourself
Same logic repeated several times

serhii-ilin: 500 is a magic number. should be either a constant or come from a config file

src/main/java/mat/client/shared/CQLLibraryResultTable.java
```diff
@@ -62,41 +64,75 @@ public Boolean getValue(CQLLibraryDataSetObject object) {
...
 			public SafeHtml getValue(CQLLibraryDataSetObject object) {
 				return CellTableUtility.getColumnToolTip(object.getVersion());
 			}
+			/*
+				Single Click to select row and enable checkbox
+				Double Click to navigate to CQL Composer tab
+			*/
+			@Override
+			public void onBrowserEvent(Cell.Context context, Element elem, CQLLibraryDataSetObject object, NativeEvent event) {
+				if (lastClick < System.currentTimeMillis() - 500)
```
serhii-ilin: The state of lastClick is shared by different instances of event handlers for different columns. can it cause any UI glitches?

adongare:
This will address all of the above issues and also no need for each column to subscribe for click event(even if we add new columns as we go) but this deals with native elements for select/deselect since `selectionModel.setSelected(obj, true/false);` not toggling the checkbox
```
table.addCellPreviewHandler(new CellPreviewEvent.Handler<ManageMeasureSearchModel.Result>() {

			@Override
			public void onCellPreview(CellPreviewEvent<ManageMeasureSearchModel.Result> event) {
				String eventType = event.getNativeEvent().getType();
				//get the native event target object
				EventTarget target = event.getNativeEvent().getCurrentEventTarget();
				//convert to native element
				Element element = Element.as(target);
				//get selected model object from event
				ManageMeasureSearchModel.Result obj = event.getValue();
				//check if event is click(this will receive all events on table)
				if (BrowserEvents.CLICK.equalsIgnoreCase(eventType)) {
					//find the checkbox Element whoch is at tbody-> tr -> td -> checkbox
					//tbody
					Element bodyElement = element.getElementsByTagName("tbody").getItem(0);
					//tr
					Element selectedRow = bodyElement.getElementsByTagName("tr").getItem(event.getIndex());
					//checkbox (no need to check for td)
					Element input = selectedRow.getElementsByTagName("input").getItem(0);
					clickCount++;
					//check if single/double click
					if (clickCount == 1) {
						//click timer
						singleClickTimer = new Timer() {
							@Override
							public void run() {
								clickCount = 0;
								InputElement checkbox = input.cast();
								//programmatically toggle checkbox
								if(checkbox.isChecked()) {
									selectionModel.setSelected(obj, false);
									selectedRow.removeClassName("GB-MJYKBBE");//to remove selection color from row
									checkbox.setChecked(false);
								} else {
									selectionModel.setSelected(obj, true);
									selectedRow.addClassName("GB-MJYKBBE");//to color row
									checkbox.setChecked(true);//GB-MJYKBPD GB-MJYKBBE GB-MJYKBGD
								}
							}
						};
						//schedule click timer
						singleClickTimer.schedule(clickDelay);
					} else if (clickCount == 2) {//double click event
						singleClickTimer.cancel();//cancel timer
						clickCount = 0;
						SelectionEvent.fire(fireEvent, obj);
					} else {
						event.setCanceled(true);
					}
				}
			}
		});
```

src/main/java/mat/client/shared/CQLLibraryResultTable.java
RohitKandimalla: you meant to say that, I have to restrict the state of lastClick for every object?

src/main/java/mat/client/shared/CQLLibraryResultTable.java
RohitKandimalla: Noted

RohitKandimalla: Event handlers over the row isn't working as expected, I have tried multiple ways to do it over the row. Though it's not the optimal way of doing it, it does its job.

serhii-ilin: In any case you don't have to repeat same code several times

PR #30, opened 2019-12-17
MAT-155 MAT UI - Add FHIR / QDM Indicator

ghost:


src/main/java/mat/client/shared/MatContext.java
```diff
@@ -87,1492 +86,1480 @@
...
-		cqlConstantService.getAllCQLConstants(new AsyncCallback<CQLConstantContainer>() {
-		
-			@Override
-			public void onFailure(Throwable caught) {
-				Window.alert(MatContext.get().getMessageDelegate().getGenericErrorMessage());
-			}
-
-			@Override
-			public void onSuccess(CQLConstantContainer result) {
-				cqlConstantContainer = result;
```
nicole-hunter: I'm a bit concerned about removing the cast to BonnieServiceAsync here. I remember it being added for a reason, I don't remember that reason. But especially since we cant test the bonnie connection locally or in dev right now I don't have warm and fuzzys removing this.

serhii-ilin: It's a false positive diff caused with sources re-formatting. Mat files have mixed identiation tab/spaces. Somehow GIT doesn't show the diffs properly. But Intellij shows it correctly. No casting was affected.

nicole-hunter (approved):

serhii-ilin:
@adongare  and @RohitKandimalla could you please look into the PR? 
It's simple ( if whitespace diffs are ignored )

adongare (approved):

PR #31, opened 2019-12-18
Feature/mat-50 & 55

adongare:
Added checkbox to select multiple measure rows in measure library page. ~~This PR is without unit tests. This is blocking MAT-51 & MAT-56. so planning to get this reviewed without unit tests and create another PR for unit test shortly.~~  Added test case to get started. Did not cover 100% but we can keep adding tests as we change something

Most recent Libs table
recentlib

My Measures table
mymeasuretable


src/main/java/mat/client/shared/MeasureLibraryResultTable.java
```diff
@@ -143,101 +149,139 @@ public void onBrowserEvent(Context context, Element elem, ManageMeasureSearchMod
...
+				@Override
+				public Result getValue(Result object) {
+					return object;
+				}
+			}, bulkExportColumnHeader);
+		}
+		// Add event handler for table
+		table.addCellPreviewHandler(event -> {
+			String eventType = event.getNativeEvent().getType();
+			EventTarget target = event.getNativeEvent().getCurrentEventTarget();
```
serhii-ilin: @RohitKandimalla - sholuld we try the same for CQL Library grids?

RohitKandimalla: I tried using addCellPreviewHandler, but couldn’t make it work.

src/main/java/mat/client/shared/MeasureLibraryResultTable.java
```diff
@@ -4,16 +4,16 @@
 import java.util.LinkedList;
 import java.util.List;
 
+import com.google.gwt.cell.client.*;
+import com.google.gwt.core.client.GWT;
+import com.google.gwt.dom.client.*;
```
serhii-ilin: @adongare  I suggest you should setup your intellij to prefer classes imports vs wild card imports

adongare: that's good idea. I'll do it. Thanks!

serhii-ilin: as I see from the MAT sources consistently adhere to classes import

serhii-ilin: let me know if you need guidance on that

src/main/java/mat/client/shared/MeasureLibraryResultTable.java
adongare: I can take a look and help if we want similar logic in CQL lib page

src/main/java/mat/client/shared/MeasureLibraryResultTable.java
adongare: I did not find any way to disable it but I increased **Class count to import with '*'** from 5 to 50. This stops wildcard imports now. remaining settings kept to default(tabs, spaces, indedents, line breaks)

src/test/java/mat/client/shared/MeasureLibraryResultTableTest.java
```diff
@@ -0,0 +1,75 @@
...
+        results = new ArrayList<>();
+        results.add(measure1);
+        results.add(measure2);
+    }
+
+    @Test
+    public void testAddColumnToTable() {
+        com.google.gwt.user.client.Element tbl = DOM.createTable();
+        when(cellTable.getElement()).thenReturn(tbl);
+        when(tbl.cast()).thenReturn(GWT.create(TableElement.class));
```
serhii-ilin: that's interesting. I haven't thought about mocking 'cast' method itself

adongare: that's lil bit strange. Not sure why mocked celltable was not getting casted to com.google.gwt.dom.client.TableElement, So had to do some mocking on getElement() & cast().

PR #32, opened 2019-12-19
Remove the lines that skip unit tests for LGTM 

casey-erdmann:
Remove the lines that skip unit tests for LGTM so they are accounted for during a build.

There is a desire to include unit test as part of the build criteria in the SAST process when LGTM builds the code. This commit modifies the lgtm.yml to not skip unit tests. While LGTM's focus is primarily security analysis, this change allows us to include successful unit tests as part as our successful build criteria which factors into our code quality over all.

This will also mean that breaking unit tests on a PR will not build successfully in LGTM. The assumed goal is that a PR should not break unit tests in any case, and if it does, it should be fixed. That said, this is an assumed case and as such should be taken into consideration during review.

Screenshot showing successful test build with new configuration was successful:
Screen Shot 2019-12-19 at 12 28 05 PM

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -4,5 +4,4 @@ extraction:
       build_command:
       - "./install-deps.sh"
       - "mvn clean package -f \"pom.xml\" -B -V -e -Dfindbugs.skip -Dcheckstyle.skip\
-        \ -Dpmd.skip=true -Denforcer.skip -Dmaven.javadoc.skip -DskipTests -Dmaven.test.skip.exec\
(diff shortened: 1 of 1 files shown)


casey-erdmann:
@serhii-ilin it should be the same simple change as far as the config goes, only caveat is testing. I'll need to test it to make sure nothing breaks/takes an extraneous amount of time on the LGTM side again. Might take me a week or so with the other items I have, but I'll add that to my list of todos for LGTM 👍

PR #33, opened 2019-12-20
Feature/mat 312 mvp feature flags

RohitKandimalla:
Added a new script to liquibase to insert new MVP Feature Flags into FEATURE_FLAGS table. All flags are set to "false" initially. "/src/main/java/liquibase/addMVPFeatureFlags.xml"

Retrieving all flags from Feature_Flag table and storing them as a list. Used inherited class GenericDAO to find all records. 

image

All Flag constants are created and could be used to know if the flag is on/off "src/main/java/mat/client/util/FeatureFlagConstant.java"

`MatContext.get().getFeatureFlagStatus(FeatureFlagConstant.MAT_ON_FHIR)` returns a Boolean value stating whether the specific flag is On/Off

### DIFF
--- a/src/main/java/mat/client/shared/MatContext.java
+++ b/src/main/java/mat/client/shared/MatContext.java
@@ -74,7 +74,6 @@ import mat.client.umls.service.VSACAPIService;
 import mat.client.umls.service.VSACAPIServiceAsync;
 import mat.client.umls.service.VsacApiResult;
 import mat.client.util.ClientConstants;
-import mat.model.FeatureFlag;
 import mat.model.GlobalCopyPasteObject;
 import mat.model.MeasureType;
 import mat.model.cql.CQLModel;
@@ -233,7 +232,7 @@ public class MatContext implements IsSerializable {
 
     private Map<String, String> expressionToReturnTypeMap = new HashMap<>();
 
-	private List<FeatureFlag> featureFlagList;
+	private Map<String, Boolean> featureFlagMap;
 
 	public void clearDVIMessages() {
         if (qdsView != null) {
@@ -1522,17 +1521,13 @@ public class MatContext implements IsSerializable {
         });
     }
 
-	public void setFeatureFlags(List<FeatureFlag> featureFlagList){
-		this.featureFlagList = featureFlagList;
+	public void setFeatureFlags(Map<String, Boolean> featureFlagMap){
+		this.featureFlagMap = featureFlagMap;
 	}
 
 	//returns if specific feature flag is on/off
 	public boolean getFeatureFlagStatus(String flag){
-		FeatureFlag featureFlag = this.featureFlagList.stream()
-				.filter(f -> flag.equals(f.getFlagName()))
-				.findAny()
-				.orElse(null);
-		return featureFlag.getFlagOn();
+		return featureFlagMap.get(flag);
 	}
 
     public List<MeasureType> getMeasureTypeList() {
--- a/src/main/java/mat/client/shared/MatContext.java
+++ b/src/main/java/mat/client/shared/MatContext.java
@@ -1527,7 +1527,7 @@ public class MatContext implements IsSerializable {
 
 	//returns if specific feature flag is on/off
 	public boolean getFeatureFlagStatus(String flag){
-		return featureFlagMap.get(flag);
+		return featureFlagMap.getOrDefault(flag,false);
 	}
 
     public List<MeasureType> getMeasureTypeList() {
--- a/src/main/java/mat/client/shared/MatContext.java
+++ b/src/main/java/mat/client/shared/MatContext.java
@@ -233,9 +233,9 @@ public class MatContext implements IsSerializable {
 
     private Map<String, String> expressionToReturnTypeMap = new HashMap<>();
 
-    private FeatureFlag matOnFHIR;
+	private List<FeatureFlag> featureFlagList;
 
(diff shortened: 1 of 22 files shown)


src/main/java/mat/client/shared/MatContext.java
```diff
@@ -233,9 +233,9 @@
 
     private Map<String, String> expressionToReturnTypeMap = new HashMap<>();
 
-    private FeatureFlag matOnFHIR;
+	private List<FeatureFlag> featureFlagList;
```
serhii-ilin: I would suggest to store a MAP here, instead of a list

src/main/java/mat/client/shared/MatContext.java
```diff
@@ -1522,13 +1522,18 @@ public void onSuccess(List<MeasureType> result) {
...
+		this.featureFlagList = featureFlagList;
+	}
 
-    public FeatureFlag getMatOnFHIR() {
-        return this.matOnFHIR;
-    }
+	//returns if specific feature flag is on/off
+	public boolean getFeatureFlagStatus(String flag){
+		FeatureFlag featureFlag = this.featureFlagList.stream()
+				.filter(f -> flag.equals(f.getFlagName()))
```
serhii-ilin: does it make sense to pre-compute a map?
or, we know for sure that the list is not going to grow

src/main/java/mat/client/util/FeatureFlagConstant.java
```diff
@@ -0,0 +1,15 @@
+package mat.client.util;
+
+public class FeatureFlagConstant {
+
+    public final static String MAT_ON_FHIR ="MAT_ON_FHIR";
```
serhii-ilin: btw,
can we use an interface for constants? it would less verbose then.
but, I don't know if our code convention allows interfaces for contants

src/main/java/mat/dao/impl/FeatureFlagDAOImpl.java
```diff
@@ -24,16 +20,8 @@ public FeatureFlagDAOImpl(@Autowired SessionFactory sessionFactory) {
...
-        final CriteriaQuery<FeatureFlag> query = cb.createQuery(FeatureFlag.class);
-        final Root<FeatureFlag> root = query.from(FeatureFlag.class);
-
-        query.select(root).where(cb.equal(root.get(FLAG_NAME), flagName));
-
-        final List<FeatureFlag> dataTypeList = session.createQuery(query).getResultList();
-
-        return CollectionUtils.isNotEmpty(dataTypeList) ? dataTypeList.get(0) : null;
+    public List<FeatureFlag> findAllFeatureFlags() {
+        final List<FeatureFlag> dataTypeList = find();
```
serhii-ilin: There is a well-known idiom: "Prefer empty collections to nulls"
http://www.javapractices.com/topic/TopicAction.do?Id=59
Should we follow it in our project?

src/test/java/mat/server/FeatureFlagServiceImplTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Mock
+    FeatureFlagDAOImpl featureFlagDAO;
+
+    @InjectMocks
+    FeatureFlagServiceImpl featureFlagServiceImpl;
+
+    @Before
+    public void init() {
+        MockitoAnnotations.initMocks(this);
```
serhii-ilin: this is not needed, since you use MockitoExtension

src/main/java/mat/dao/impl/FeatureFlagDAOImpl.java
serhii-ilin: I suggest that Dao implementation should retrieve a list of FeatureFlags and then transform into a map.

src/main/java/mat/server/FeatureFlagServiceImpl.java
```diff
@@ -8,17 +8,19 @@
...
 	private static final long serialVersionUID = 1L;
 
 	@Autowired
 	FeatureFlagDAO featureFlagDAO;
-	
+
 	@Override
-	public FeatureFlag findFeatureFlag(String flagName) {
-		return featureFlagDAO.findFlagByName(flagName);
+	public List<FeatureFlag> findFeatureFlag() {
```
serhii-ilin: it probably makes sense to give it a plural name, like findFeagureFlags or findAllFeatureFlags or something similar

RohitKandimalla: Noted

src/test/java/mat/server/FeatureFlagServiceImplTest.java
RohitKandimalla: Initialization is removed

src/main/java/mat/client/util/FeatureFlagConstant.java
RohitKandimalla: I used Interface for Constants

src/main/java/mat/client/shared/MatContext.java
RohitKandimalla: Changed it to MAP

src/main/java/mat/dao/impl/FeatureFlagDAOImpl.java
RohitKandimalla: Updated

PR #34, opened 2019-12-23
MAT-51 and MAT-56 measures grid toolbar

ghost:
The change covers MAT-51 and MAT-56 and bug MAT-406

image


serhii-ilin:
The change covers MAT-51 and MAT-56 and bug MAT-406

src/main/java/mat/client/measure/ManageMeasureSearchView.java
```diff
@@ -36,435 +32,352 @@
...
+
+    private final MessageAlert errorMessages = new ErrorMessageAlert();
+
+    private final MessageAlert errorMessagesForBulkExport = new ErrorMessageAlert();
+
+    private ManageMeasureSearchModel data = new ManageMeasureSearchModel();
+
+    private final FormPanel form = new FormPanel();
+
+//    public VerticalPanel measureVpanel = new VerticalPaGrid508nel();
```
nicole-hunter: This looks like we have commented it out. Do we still need it?

src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
```diff
@@ -0,0 +1,163 @@
...
+        exportButton.setText("Export");
+        exportButton.setTitle("Click to export");
+        exportButton.setWidth("72px");
+        buildActionButton(exportButton, IconType.DOWNLOAD);
+    }
+
+    private void applyDefaultAllButExport() {
+        versionButton.setText("Create Version/Draft");
+        versionButton.setTitle("Click to create version or draft");
+        versionButton.setWidth("146px");
```
adongare: Maybe all these 3 line can go into buildActionButton(...)

src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
```diff
@@ -0,0 +1,163 @@
...
+        cloneButton.setWidth("69px");
+        buildActionButton(cloneButton, IconType.CLONE);
+    }
+
+    private Button buildActionButton(Button actionButton, IconType icon) {
+        actionButton.setType(ButtonType.DEFAULT);
+        actionButton.setSize(ButtonSize.SMALL);
+        actionButton.setIcon(icon);
+        actionButton.setEnabled(false);
+        return actionButton;
```
adongare: doesn't use returned value anywhere. Can this have void return type?

war/Mat.css
```diff
@@ -2956,3 +2958,11 @@ table#MostRecentActivityCellTable {
 .table td {
     vertical-align: middle;
 }
+
+.action-button-bar {
+}
```
adongare: I think empty not needed

war/Mat.css
```diff
@@ -2956,3 +2958,11 @@ table#MostRecentActivityCellTable {
 .table td {
     vertical-align: middle;
 }
+
+.action-button-bar {
```
nicole-hunter: Do we need this blank style?

src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
```diff
@@ -0,0 +1,163 @@
...
+        buildActionButton(exportButton, IconType.DOWNLOAD);
+    }
+
+    private void applyDefaultAllButExport() {
+        versionButton.setText("Create Version/Draft");
+        versionButton.setTitle("Click to create version or draft");
+        versionButton.setWidth("146px");
+        buildActionButton(versionButton, IconType.STAR);
+
+        historyButton.setText("History");
```
adongare: I think this label should be **View History**

src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
```diff
@@ -0,0 +1,163 @@
...
+        shareButton = GWT.create(Button.class);
+        cloneButton = GWT.create(Button.class);
+        exportButton = GWT.create(Button.class);
+
+        add(versionButton);
+        add(historyButton);
+        add(editButton);
+        add(shareButton);
+        add(cloneButton);
+        add(exportButton);
```
adongare: Not sure if the order of the buttons is fine. Wireframe shows different order

adongare (approved):
Looks really good to me!

src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
serhii-ilin: yes, you're right

war/Mat.css
serhii-ilin: Yeah. I was thinking to keep an empty style as a placeholder for a common style for all instances of toolbars. Do you think we don't need it?

src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
serhii-ilin: I changed it to History so it's consistent and looks the same as the CQL toolbar. I will clarify it with Kathy

src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
serhii-ilin: I changed it to resemble the CQL toolbar. I will clarify it with Kathy

war/Mat.css
serhii-ilin: Not that we really need it. I was thinking to keep an empty style as a placeholder for a common style for all instances of toolbars. Do you think we don't need it?

src/main/java/mat/client/measure/ManageMeasureSearchView.java
serhii-ilin: Ups, I missed that. Will remove

src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
serhii-ilin: Yeah, I hesitated because of the number of arguments. 5 args is quite high. But there are no other good options. I will move it to the method.

war/Mat.css
serhii-ilin: In addition to that, the style is used as a classifier for the buttons. So all toolbar buttons have same padding.

war/Mat.css
serhii-ilin: In addition to that, the style is used as a classifier for the buttons. So all toolbar buttons have same padding.

nicole-hunter (approved):

PR #36, opened 2019-12-26
Feature/mat 394 jdk11

ghost:
The change was done in order to asses migration to JDK 11.

Following libraries were upgraded: spring, spring boot, spring-security, hibernate, jpa, javax, javaassist, jdk-11 and some others.

GWT framework doesn't currently support sourceLevel 11 (it's supported in GWT HEAD, but not released yet if it's ever going to be released). 

Thus gwt sourceLevel is set to 1.8. Not sure what issues can it cause, but visually the app is working properly. GWT super dev debugging mode works fine as well.


src/main/java/mat/server/Application.java
```diff
@@ -117,4 +125,41 @@ public OTPValidatorInterfaceForUser matOtpValidator(@Value("${2FA_AUTH_CLASS}")
...
+                .and()
+            .sessionManagement()
+                .invalidSessionUrl("/Login.html")
+                .maximumSessions(1);
+    }
+
+    @Override
+    protected UserDetailsService userDetailsService() {
+        UserDetails user =
+                User.withDefaultPasswordEncoder()
```
serhii-ilin: I preserved that original "dude" user :) Most likely we can drop it completely.

AndrewBird81 (approved):
LGTM

adongare (approved):

PR #38, opened 2019-12-30
MAT-354 MAT - FHIR Validation Report

adongare:
Implemented FHIR conversion validation report with mock data. A new story will be created to handle following pending things:
1. Once we have conversion service deployed, need to revisit and replace mock data call with call to actual service. 
2. There is another story to implement "Validate FHIR Conversion" button in UI which is "in progress". Once it is merged need to link button click action.
3. `Fields to reconcile` section from mockup is not yet available in validation service. One it is available, report will have once more section to display `Fields to reconcile` section 

FHIR validation report


src/main/java/mat/server/fhirvalidation/FhirValidationReportServlet.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.context.ApplicationContext;
+import org.springframework.http.MediaType;
+import org.springframework.web.context.support.WebApplicationContextUtils;
+
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class FhirValidationReportServlet extends HttpServlet {
```
serhii-ilin: Since MAT is a Spring WebMVC application we might prefer to go with a "@Controller" instead of a plain servlet. It should be a small change

src/main/java/mat/server/fhirvalidation/FhirValidationReportServlet.java
```diff
@@ -0,0 +1,49 @@
...
+        applicationContext = WebApplicationContextUtils.getRequiredWebApplicationContext(getServletContext());
+        String id = req.getParameter(ID_PARAM);
+        logger.info("Generating validation report for measure id: " + id);
+        FhirValidationReportService fhirValidationReportService = getValidationReportService();
+        String validationReport = "";
+        try {
+            validationReport = fhirValidationReportService.getFhirConversionReportForMeasure(id);
+        } catch (TemplateException e) {
+            logger.error("Exception occurred while generation FHIR conversion report:", e);
+            validationReport = "Error occurred while validating the FHIR conversion, please try after some time";
```
serhii-ilin: do you want to set HTTP status code?

src/main/java/mat/server/fhirvalidation/FhirValidationReportServlet.java
adongare: Right now I don't see WebMvc enabled. Do you want me to add configuration for this and modify this accordingly?

src/main/java/mat/server/fhirvalidation/FhirValidationReportServlet.java
adongare: I'll set it, thanks

src/main/java/mat/server/fhirvalidation/FhirValidationReportServlet.java
serhii-ilin: Right. I was under impression SpingMVC is fully configured. We will have to re-visit it later..

PR #39, opened 2019-12-30
Feature/mat 353 run fhir validation button

RohitKandimalla:
Added Run FHIR Validation Button for both Measure Grids.

image

Button will be enabled/disabled based on following conditions.

image

### DIFF
--- a/src/main/java/mat/client/measure/ManageMeasurePresenter.java
+++ b/src/main/java/mat/client/measure/ManageMeasurePresenter.java
@@ -1459,6 +1459,11 @@ public class ManageMeasurePresenter implements MatPresenter, TabObserver {
             }
 
             @Override
+            public void onFhirValidationClicked(ManageMeasureSearchModel.Result result) {
+                //TODO Calling MAT-354 service
+            }
+
+            @Override
             public void onShareClicked(ManageMeasureSearchModel.Result result) {
                 resetMeasureFlags();
                 displayShare(null, result.getId(), result.getName());
--- a/src/main/java/mat/client/measure/ManageMeasureSearchModel.java
+++ b/src/main/java/mat/client/measure/ManageMeasureSearchModel.java
@@ -154,6 +154,8 @@ public class ManageMeasureSearchModel implements IsSerializable, SearchResults<M
 
         private String cqlLibraryName;
 
+        private boolean isValidatable;
+
         private int clickCount = 0;
 
         public Result() {
@@ -521,6 +523,24 @@ public class ManageMeasureSearchModel implements IsSerializable, SearchResults<M
          */
         public void setVersion(String version) {
             this.version = version;
+        }
+
+        /**
+         * Gets the isValidatable.
+         *
+         * @return the isValidatable
+         */
+        public boolean isValidatable() {
+            return this.isValidatable;
+        }
+
+        /**
+         * Sets the isValidatable.
+         *
+         * @param isValidatable the new isValidatable
+         */
+        public void setValidatable(boolean isValidatable) {
+            this.isValidatable = isValidatable;
         }
 
         /**
--- a/src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
+++ b/src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
@@ -1,7 +1,8 @@
 package mat.client.shared;
 
 import java.util.Collection;
-
+import com.google.gwt.regexp.shared.MatchResult;
+import com.google.gwt.regexp.shared.RegExp;
 import org.gwtbootstrap3.client.ui.Button;
 import org.gwtbootstrap3.client.ui.constants.ButtonSize;
 import org.gwtbootstrap3.client.ui.constants.ButtonType;
@@ -19,6 +20,10 @@ public class MeasureLibraryGridToolbar extends HorizontalFlowPanel {
     private Button shareButton;
     private Button cloneButton;
     private Button exportButton;
+    private Button fhirValidationButton;
+    private float version;
+
+    private static final float RUN_FHIR_VALIDATION_VERSION = 5.8F;
 
     public MeasureLibraryGridToolbar() {
         setStyleName("action-button-bar");
@@ -31,6 +36,7 @@ public class MeasureLibraryGridToolbar extends HorizontalFlowPanel {
         shareButton = GWT.create(Button.class);
         cloneButton = GWT.create(Button.class);
         exportButton = GWT.create(Button.class);
+        fhirValidationButton = GWT.create(Button.class);
 
         add(versionButton);
         add(historyButton);
@@ -38,7 +44,7 @@ public class MeasureLibraryGridToolbar extends HorizontalFlowPanel {
         add(shareButton);
         add(cloneButton);
         add(exportButton);
-
+        add(fhirValidationButton);
 
         applyDefault();
     }
@@ -55,6 +61,7 @@ public class MeasureLibraryGridToolbar extends HorizontalFlowPanel {
         buildButton(editButton, IconType.EDIT, "Edit", "Click to edit", "57px");
         buildButton(shareButton, IconType.SHARE_SQUARE, "Share", "Click to share", "68px");
         buildButton(cloneButton, IconType.CLONE, "Clone", "Click to clone", "69px");
+        buildButton(fhirValidationButton, IconType.FILE_TEXT_O, "Run FHIR Validation", "Click to Run FHIR Validation", "146px");
     }
 
     private void buildButton(Button actionButton, IconType icon, String text, String title, String width) {
@@ -89,6 +96,21 @@ public class MeasureLibraryGridToolbar extends HorizontalFlowPanel {
         }
 
         historyButton.setEnabled(true);
+
+        RegExp regExp = RegExp.compile("[0-9]+\\.[0-9]+");
+        MatchResult matcher = regExp.exec(selectedItem.getVersion());
+        boolean matchFound = matcher != null;
+        if (matchFound) {
+            version = Float.parseFloat(matcher.getGroup(0));
+        }
+
+        if( (selectedItem.isDraft() && selectedItem.getMeasureModel().equals("FHIR")) ||
+                (selectedItem.isDraftable() && selectedItem.getMeasureModel().equals("QDM") && version > RUN_FHIR_VALIDATION_VERSION) ){
+            fhirValidationButton.setEnabled(true);
+        }else{
+            fhirValidationButton.setEnabled(false);
+        }
+
 
         if (selectedItem.isEditable()) {
             if (selectedItem.isMeasureLocked()) {
@@ -137,6 +159,10 @@ public class MeasureLibraryGridToolbar extends HorizontalFlowPanel {
 
     public Button getExportButton() {
         return exportButton;
+    }
+
+    public Button getFhirValidationButton(){
+        return fhirValidationButton;
     }
 
 }
--- a/src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
+++ b/src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
@@ -55,7 +55,7 @@ public class MeasureLibraryGridToolbar extends HorizontalFlowPanel {
         buildButton(exportButton, IconType.DOWNLOAD, "Export", "Click to export", "72px");
     }
 
-    public void addFhirValidationButton(){
+    public void addFhirValidationButton() {
         if (MatContext.get().getFeatureFlagStatus(FeatureFlagConstant.FHIR_CONV_V1)) {
             add(fhirValidationButton);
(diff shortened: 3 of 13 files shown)


src/test/java/mat/client/shared/MeasureLibraryGridToolbarTest.java
```diff
@@ -109,6 +110,15 @@ public void testExportOnEmptySelection() {
         Mockito.verify(toolbar.getExportButton(), Mockito.times(1)).setWidth(Mockito.eq("72px"));
     }
 
+    public void testApplyDefaultFhirValidationButton() {
```
serhii-ilin: Looks like the test is not annotated. I will not run

src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
```diff
@@ -49,6 +55,13 @@ void applyDefault() {
         buildButton(exportButton, IconType.DOWNLOAD, "Export", "Click to export", "72px");
     }
 
+    public void addFhirValidationButton(){
+        if (MatContext.get().getFeatureFlagStatus(FeatureFlagConstant.FHIR_CONV_V1)) {
```
serhii-ilin: why cannot we call it from appluDefault() ?
what is the aim of a separate public method call?

src/test/java/mat/client/shared/MeasureLibraryGridToolbarTest.java
RohitKandimalla: Added it

src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
```diff
@@ -90,6 +103,23 @@ public void updateOnSelectionChanged(Collection<ManageMeasureSearchModel.Result>
 
         historyButton.setEnabled(true);
 
+        //Run FHIR validation Button
```
serhii-ilin: 1. **Responsibility:** I think it's not the responsibility of a toolbar to do this business logic.

See how selectedItem.isSharable() or selectedItem.isExportable() are implemented. In a similar way you can introuce selectedItem.isValidatable()

2. RegExp parsing is more expensive and verbose then if you parse versions as BigDecimal.

src/main/java/mat/client/shared/MeasureLibraryResultTable.java
```diff
@@ -197,6 +201,15 @@ void onVersionButtonClicked(MultiSelectionModel<Result> selectionModel) {
         });
     }
 
+    @VisibleForTesting
+    void onFhirValidationButtonClicked (MultiSelectionModel < Result > selectionModel) {
```
serhii-ilin: line formatting is not OK

src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
RohitKandimalla: Test cases are failing if the getFeatureFlagStatus is called from constructor (applyDefault) because of toolbar being Injected prior to that. 

image

src/main/java/mat/client/measure/ManageMeasurePresenter.java
```diff
@@ -1458,6 +1458,11 @@ public void onHistoryClicked(ManageMeasureSearchModel.Result result) {
                 displayHistory(result.getId(), result.getName());
             }
 
+            @Override
+            public void onFhirValidationClicked(ManageMeasureSearchModel.Result result) {
+                //TODO Calling MAT-354 service
```
serhii-ilin: MAT-354 is in review already

src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
```diff
@@ -19,6 +21,10 @@
     private Button shareButton;
     private Button cloneButton;
     private Button exportButton;
+    private Button fhirValidationButton;
+    private float version;
+
+    private static final float RUN_FHIR_VALIDATION_VERSION = 5.8F;
```
serhii-ilin: 1. BigDecimal is a better option.
2. The toolbar should not be aware of validation eligibility details

src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
```diff
@@ -19,6 +21,10 @@
     private Button shareButton;
     private Button cloneButton;
     private Button exportButton;
+    private Button fhirValidationButton;
+    private float version;
```
serhii-ilin: Is version field used somewhere?

src/main/java/mat/client/shared/MeasureLibraryResultTable.java
RohitKandimalla: Modified

src/main/java/mat/client/measure/ManageMeasurePresenter.java
RohitKandimalla: There is an another small story which implements this part.

src/main/java/mat/client/shared/MeasureLibraryGridToolbar.java
RohitKandimalla: yes

src/main/java/mat/server/MeasureLibraryServiceImpl.java
```diff
@@ -1138,7 +1139,7 @@ public ManageMeasureSearchModel getAllRecentMeasureForUser(String userId) {
             lockedUserInfo.setLastName(measure.getLockedUser().getLastName());
             detail.setLockedUserInfo(lockedUserInfo);
         }
-
+        detail.setValidatable(MeasureDetailsUtil.isValidatable(measure.getReleaseVersion(), measure.isDraft(), detail.isDraftable(), measure.getMeasureModel()));
```
adongare: I would pass in measure object to `isValidatable(measure)`

serhii-ilin: Yeah. I agree. It makes sense to mass measure object. Also, the pattern for other similar methods MeasureDetailsUtil.get().isValidatable

src/main/java/mat/shared/model/util/MeasureDetailsUtil.java
```diff
@@ -94,4 +99,18 @@ public static String defaultTypeIfBlank(String type) {
         return type == null || type.isEmpty() ? PRE_CQL : type;
     }
 
+    public static Boolean isValidatable(String releaseVersion, boolean draft, boolean draftable, String modelType){
```
serhii-ilin: should be primitive boolean

src/main/java/mat/shared/model/util/MeasureDetailsUtil.java
```diff
@@ -94,4 +99,18 @@ public static String defaultTypeIfBlank(String type) {
...
+        BigDecimal version = null;
+        RegExp regExp = RegExp.compile("[0-9]+\\.[0-9]+");
+        MatchResult matcher = regExp.exec(releaseVersion);
+        if (matcher != null) {
+            version = new BigDecimal(matcher.getGroup(0));
+        }
+        if (modelType != null && ((draft && modelType.equals(FHIR)) ||
```
serhii-ilin: you can return the expression value, 
instead of 
    if (expression) { return true; } else {return false}
it would be
    return expression;

src/main/java/mat/shared/model/util/MeasureDetailsUtil.java
RohitKandimalla: Modified

src/main/java/mat/shared/model/util/MeasureDetailsUtil.java
RohitKandimalla: that would be repetitive of same expression right ?

src/main/java/mat/server/MeasureLibraryServiceImpl.java
RohitKandimalla: But one of the parameter we need is isDraftable, which is not present in measure Entity. isDraftable is decided based on isDraft(). But @serhii-ilin you told the other day that its not a good practice to use isDraft() to decide if it is Draftable. 

isValidatable() is being called for recent activity grid and measure grid, though both the calls are from same class, they use 2 different Models(measure & dto).
`detail.setValidatable(MeasureDetailsUtil.isValidatable(dto.getReleaseVersion(), dto.getQdmVersion(), dto.isDraft(), dto.isDraftable(), dto.getMeasureModel()));
`

src/main/java/mat/shared/model/util/MeasureDetailsUtil.java
serhii-ilin: why repetitive? currently you just return true/false. It's not a good practice..

src/main/java/mat/client/measure/ManageMeasureSearchModel.java
```diff
@@ -523,6 +525,24 @@ public void setVersion(String version) {
...
+         * @return the isValidatable
+         */
+        public boolean isValidatable() {
+            return this.isValidatable;
+        }
+
+        /**
+         * Sets the isValidatable.
+         *
+         * @param isValidatable the new isValidatable
```
serhii-ilin: I think you missed equals/hashCode methods. I don't know why those are needed in the MAT App.. but their implementation is based on the fields/flags of the object.

src/main/java/mat/model/clause/MeasureShareDTO.java
```diff
@@ -418,6 +420,25 @@ public String getReleaseVersion() {
...
+	 *
+	 * @return the qdmVersion
+	 */
+	public String getQdmVersion() {
+		return qdmVersion;
+	}
+
+	/**
+	 * Sets the qdmVersion.
+	 *
```
serhii-ilin: I believe we don't need auto-generated comments. It just clutters the code.

RohitKandimalla: Do I have to remove comments for entire MeasureShareDTO class ?

src/main/java/mat/shared/model/util/MeasureDetailsUtil.java
RohitKandimalla: Updated

src/main/java/mat/server/MeasureLibraryServiceImpl.java
serhii-ilin: I think there is a misunderstanding... It's not what I meant. Will ping you on that

src/main/java/mat/model/clause/MeasureShareDTO.java
serhii-ilin: I believe so. Why would you need any pointless comments.. 

https://www.todaysoftmag.com/article/1120/clean-code-comments-and-formatting

RohitKandimalla: Done

src/main/java/mat/shared/model/util/MeasureDetailsUtil.java
```diff
@@ -94,4 +100,22 @@ public static String defaultTypeIfBlank(String type) {
...
     }
 
+    public static boolean isValidatable(String releaseVersion, String qdmVersion, boolean draft, boolean draftable, String modelType) {
+        if (releaseVersion == null || qdmVersion == null || modelType == null) {
+            return false;
+        }
+        BigDecimal matVersion = null;
+        BigDecimal measureVersion = new BigDecimal(qdmVersion);
+
+        RegExp regExp = RegExp.compile("[0-9]+\\.[0-9]+");
```
serhii-ilin: regexp compilation is considered to be expensive. you might want to cache it in a static field.

src/main/java/mat/shared/model/util/MeasureDetailsUtil.java
```diff
@@ -94,4 +100,22 @@ public static String defaultTypeIfBlank(String type) {
...
+        MatchResult matcher = regExp.exec(releaseVersion);
+        if (matcher != null) {
+            matVersion = new BigDecimal(matcher.getGroup(0));
+        }
+
+        return (draft && modelType.equals(FHIR))
+                || (draftable && modelType.equals(QDM)
+                && RUN_FHIR_VALIDATION_VERSION.compareTo(matVersion) == -1
```
serhii-ilin: Interesting.. it took me quite a while to realize why ==-1 works for you and not >= 0.
Apparently you use 5.7 and 5.4 instead of 5.8 and 5.5 :)

RohitKandimalla: Do I have to use <= 0 instead of -1 ?

serhii-ilin: Not sure about that. Your approach is valid too.

RohitKandimalla: Using <= 0 instead of -1

PR #42, opened 2020-01-02
Feature/mat 352 mvp mat add convert toolbar

ghost:
image

As a measure developer, I want to be able to take an action in the MAT so that I can convert a QDM measure to FHIR.

In this story, we will add a Convert to FHIR toolbar button to both MAT measure grids.

The mockup for the UI is found here:  https://balsamiq.cloud/sg54b2t/pt8orhx/rDBB8

Acceptance Criteria:

0.0 This is behind the FhirConvV1 feature flag.

1.0 The Convert to FHIR toolbar button is enabled for versioned QDM (CQL) measures

     1.1 The button is disabled for draft, Pre-CQL measures and Fhir Measures 

     1.2 The button is disabled if the QDM version is before 5.5 or the MAT version is before 5.8.

2.0 If user selects a QDM (CQL) measure that has already been converted and the associated FHIR measure still exists display an 'Are you sure' prompt

     2.1 Issue the message as a dialog box. Prompt text: "Are you sure you want to convert this measure again? The existing FHIR measure will be overwritten."  Provide a "Yes" button and a "No" button

    2.2 If the user clicks Yes, we'll close the dialog box.    In a future story, when the user answers Yes, we will proceed with the conversion as outlined in MAT-114 The existing (previously converted) FHIR measure will be overwritten.  Note: this will be done in MAT-352

    2.3 If the user clicks No, close the dialog box and do nothing.

 

3.0 The Convert to FHIR button is added to the My/All Measures grid in the Measure Library page.

4.0 The Convert to FHIR button is added to the Recent Actvity grid in the Measure Library page.

5.0 The UI adheres to the mockups provided.


src/main/java/mat/client/measure/ManageMeasurePresenter.java
```diff
@@ -1508,9 +1508,43 @@ public void onSuccess(ManageMeasureDetailModel result) {
...
+                });
+
+                confirmationDialogBox.show();
+            }
+
         };
     }
 
+    private void executeConversion(Result object) {
+        Window.alert("Please wait. Conversion is in progress..");
```
adongare: is this needed?

adongare (approved):
LGTM

src/main/java/mat/client/measure/ManageMeasurePresenter.java
serhii-ilin: Nop. Just a stub for the real service

nicole-hunter (approved):

PR #44, opened 2020-01-06
Feature/mat 316 implement fhir edit feature flag

RohitKandimalla:
Acceptance Criteria:

1. The FhirEdit feature flag is currently off in Dev and the sandbox environment. When we begin the edit work, the flag will be on in Dev.
2. When the flag is off, the Edit button is disabled for Fhir Measures. It is enabled for QDM measures in a draft state.
3. When the flag is off, the double click of a measure grid row does not launch the edit / view of a measure if the measure is a fhir measure but does for a QDM measure.
4. When the flag is on, the Edit button is enabled for Fhir or QDM Measures, per the edit rules. (enabled only for measures in a draft state. Enabled only for the measure owner or those whom the measure was shared with. Otherwise disabled.)
5. This impacts both measure library grids.
6. This impacts both cql grids.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -500,9 +500,8 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.springframework</groupId>
-            <artifactId>spring-context</artifactId>
-            <version>5.2.2.RELEASE</version>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-cache</artifactId>
         </dependency>
     </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -499,6 +499,11 @@
             <version>0.8.13</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-context</artifactId>
+            <version>5.2.2.RELEASE</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/src/main/java/mat/client/measure/ManageMeasureSearchModel.java
+++ b/src/main/java/mat/client/measure/ManageMeasureSearchModel.java
@@ -7,8 +7,11 @@ import java.util.Objects;
 
 import com.google.gwt.user.client.rpc.IsSerializable;
 import com.google.gwt.user.client.ui.Widget;
+import mat.client.shared.MatContext;
 import mat.client.shared.search.SearchResults;
+import mat.client.util.FeatureFlagConstant;
 import mat.model.LockedUserInfo;
+import mat.shared.model.util.MeasureDetailsUtil;
 
 /**
  * The Class ManageMeasureSearchModel.
@@ -353,6 +356,9 @@ public class ManageMeasureSearchModel implements IsSerializable, SearchResults<M
             this.isEditable = isEditable;
         }
 
+        public boolean isFhirEditableFeatureFlag() {
+            return MatContext.get().getFeatureFlagStatus(FeatureFlagConstant.FHIR_EDIT) || (this.draft && MeasureDetailsUtil.QDM.equals(this.measureModel));
+        }
         /**
          * Checks if is clonable.
          *
--- a/src/main/java/mat/client/measure/ManageMeasureSearchModel.java
+++ b/src/main/java/mat/client/measure/ManageMeasureSearchModel.java
@@ -77,6 +77,8 @@ public class ManageMeasureSearchModel implements IsSerializable, SearchResults<M
 
         private int clickCount = 0;
 
+        private boolean isFhirEditorViewable;
+
         public Result() {
         }
 
@@ -720,6 +722,15 @@ public class ManageMeasureSearchModel implements IsSerializable, SearchResults<M
         public void incrementClickCount() {
             this.clickCount++;
         }
+
+        public void setFhirEditOrViewable(boolean isFhirEditorViewable) {
+            this.isFhirEditorViewable = isFhirEditorViewable;
+        }
+
+        public boolean isFhirEditOrViewable() {
+            return this.isFhirEditorViewable;
+        }
+
     }
 
     /**
--- a/src/main/java/mat/client/measure/ManageMeasureSearchModel.java
+++ b/src/main/java/mat/client/measure/ManageMeasureSearchModel.java
@@ -356,9 +356,6 @@ public class ManageMeasureSearchModel implements IsSerializable, SearchResults<M
             this.isEditable = isEditable;
         }
 
-        public boolean isFhirEditableFeatureFlag() {
-            return MatContext.get().getFeatureFlagStatus(FeatureFlagConstant.FHIR_EDIT) || (this.draft && MeasureDetailsUtil.QDM.equals(this.measureModel));
-        }
         /**
          * Checks if is clonable.
          *
--- a/src/main/java/mat/client/shared/CQLLibraryResultTable.java
+++ b/src/main/java/mat/client/shared/CQLLibraryResultTable.java
@@ -66,7 +66,7 @@ public class CQLLibraryResultTable {
                 if (object.getLastClick() < System.currentTimeMillis() - DELAY_TIME) {
                     object.setSelected(!object.isSelected());
                     selectionModel.setSelected(object, object.isSelected());
-                } else {
+                } else if (object.isFhirEditableFeatureFlag()) {
                     SelectionEvent.fire(fireEvent, object);
                 }
                 object.setLastClick(System.currentTimeMillis());
@@ -92,7 +92,7 @@ public class CQLLibraryResultTable {
                 if (object.getLastClick() < System.currentTimeMillis() - DELAY_TIME) {
                     object.setSelected(!object.isSelected());
                     selectionModel.setSelected(object, object.isSelected());
-                } else {
+                } else if (object.isFhirEditableFeatureFlag()) {
                     SelectionEvent.fire(fireEvent, object);
                 }
                 object.setLastClick(System.currentTimeMillis());
@@ -117,7 +117,7 @@ public class CQLLibraryResultTable {
                 if (object.getLastClick() < System.currentTimeMillis() - DELAY_TIME) {
                     object.setSelected(!object.isSelected());
                     selectionModel.setSelected(object, object.isSelected());
-                } else {
+                } else if (object.isFhirEditableFeatureFlag()) {
                     SelectionEvent.fire(fireEvent, object);
                 }
                 object.setLastClick(System.currentTimeMillis());
--- a/src/main/java/mat/client/shared/CQLLibraryResultTable.java
+++ b/src/main/java/mat/client/shared/CQLLibraryResultTable.java
@@ -66,7 +66,7 @@ public class CQLLibraryResultTable {
                 if (object.getLastClick() < System.currentTimeMillis() - DELAY_TIME) {
                     object.setSelected(!object.isSelected());
                     selectionModel.setSelected(object, object.isSelected());
-                } else if (object.isFhirEditableFeatureFlag()) {
+                } else if (object.isEditable()) {
                     SelectionEvent.fire(fireEvent, object);
                 }
                 object.setLastClick(System.currentTimeMillis());
@@ -92,7 +92,7 @@ public class CQLLibraryResultTable {
                 if (object.getLastClick() < System.currentTimeMillis() - DELAY_TIME) {
                     object.setSelected(!object.isSelected());
                     selectionModel.setSelected(object, object.isSelected());
-                } else if (object.isFhirEditableFeatureFlag()) {
+                } else if (object.isEditable()) {
                     SelectionEvent.fire(fireEvent, object);
                 }
                 object.setLastClick(System.currentTimeMillis());
@@ -117,7 +117,7 @@ public class CQLLibraryResultTable {
                 if (object.getLastClick() < System.currentTimeMillis() - DELAY_TIME) {
                     object.setSelected(!object.isSelected());
                     selectionModel.setSelected(object, object.isSelected());
-                } else if (object.isFhirEditableFeatureFlag()) {
+                } else if (object.isEditable()) {
                     SelectionEvent.fire(fireEvent, object);
                 }
                 object.setLastClick(System.currentTimeMillis());
--- a/src/main/java/mat/client/shared/CQLLibraryResultTable.java
+++ b/src/main/java/mat/client/shared/CQLLibraryResultTable.java
@@ -66,7 +66,7 @@ public class CQLLibraryResultTable {
                 if (object.getLastClick() < System.currentTimeMillis() - DELAY_TIME) {
                     object.setSelected(!object.isSelected());
                     selectionModel.setSelected(object, object.isSelected());
-                } else if (object.isEditable()) {
+                } else if (object.isFhirEditOrViewable()) {
                     SelectionEvent.fire(fireEvent, object);
                 }
                 object.setLastClick(System.currentTimeMillis());
@@ -92,7 +92,7 @@ public class CQLLibraryResultTable {
                 if (object.getLastClick() < System.currentTimeMillis() - DELAY_TIME) {
                     object.setSelected(!object.isSelected());
                     selectionModel.setSelected(object, object.isSelected());
-                } else if (object.isEditable()) {
+                } else if (object.isFhirEditOrViewable()) {
                     SelectionEvent.fire(fireEvent, object);
                 }
                 object.setLastClick(System.currentTimeMillis());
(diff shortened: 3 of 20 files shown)


src/main/java/mat/client/shared/MeasureLibraryResultTable.java
```diff
@@ -104,7 +104,7 @@ public void run() {
                         }
                     };
                     singleClickTimer.schedule(MOUSE_CLICK_DELAY);
-                } else if (obj.getClickCount() == 2) {
+                } else if (obj.getClickCount() == 2 && obj.isFhirEditableFeatureFlag()) {
```
serhii-ilin: 1. My concern that the notion of feature-flags penetrates all aspects of the business logic...It makes our UI and DTO objects coupled with both FLAGS and FHIR (https://enterprisecraftsmanship.com/posts/cohesion-coupling-difference/).
I would suggest using 'isEditable'  instead. Then 'isEditable' / isVieweable can be calculated based on flags and model type. So we will minimize future code changes..

See https://enterprisecraftsmanship.com/posts/cohesion-coupling-difference/
2. Some of the items cannot be edited but can be viewed in read-only mode. The same handlers are used to view items. Current logic can possibly block it. Could you please double check it's not blocked?

RohitKandimalla: At first I tried to use existing 'isEditable' to do calculations based on flags and model type. But Mat.Context.get().getFeatureFlagStatus() isn't available in server side. When I tried to access MatContext in isCurrentMeasureEditable(). It is throwing null pointer.

serhii-ilin: Well, feature flags are accessible through FeatureFlagService on the server-side.
Also, we might want to add caching (spring cache or hibernate caching) and optionally feature flag facade.

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
```diff
@@ -11,17 +14,32 @@
...
     private static BigDecimal QDM_VER_BEFORE_CONVERSION = new BigDecimal("5.5");
     private static BigDecimal MAT_VER_BEFORE_CONVERSION = new BigDecimal("5.8");
 
     private boolean isMeasure;
 
-    private static MatContextServiceUtil instance = new MatContextServiceUtil();
+    private static MatContextServiceUtil instance;
+
+    @Autowired
+    FeatureFlagServiceImpl featureFlag;
```
serhii-ilin: can we inject it with an interface?

src/main/java/mat/client/shared/CQLLibraryResultTable.java
```diff
@@ -117,7 +117,7 @@ public void onBrowserEvent(Cell.Context context, Element elem, CQLLibraryDataSet
                 if (object.getLastClick() < System.currentTimeMillis() - DELAY_TIME) {
                     object.setSelected(!object.isSelected());
                     selectionModel.setSelected(object, object.isSelected());
-                } else {
+                } else if (object.isEditable()) {
```
serhii-ilin: as I remember some of read-only items should be clickable/vieweable. How is this compatible with isEditable() ?

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
RohitKandimalla: yup I did it

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
```diff
@@ -11,17 +14,32 @@
...
+
+    @Override
+    public void afterPropertiesSet() {
+        instance = this;
+    }
+
+    private Map<String, Boolean> featureFlagMap = new HashMap<>();
```
serhii-ilin: why do you store it here? if it's a singleton then it's not thread-safe.
also, it's not for caching because it's always refreshed..
could you explain?

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
```diff
@@ -206,4 +233,13 @@ private BigDecimal asDecimalVersion(String version) {
         }
     }
 
+    public boolean isMeasureEditable(Measure measure) {
+        featureFlagMap = featureFlag.findFeatureFlags();
+        return featureFlagMap.get(FeatureFlagConstant.FHIR_EDIT) || MeasureDetailsUtil.QDM.equals(measure.getMeasureModel());
```
serhii-ilin: getOrDefault(..., false) would be a better way to handle it

src/main/java/mat/client/shared/CQLLibraryResultTable.java
RohitKandimalla: Double Click works similar to edit button in grid toolbar.

serhii-ilin: yeah, I just wanted to dobule check that it's not breaking anything
1. is the item editable to allow double click? What about isViewable read-only times?
2. is the change to isEditable affecting how toolbar's 'Edit' button enable/disable logic works?

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
```diff
@@ -206,4 +233,13 @@ private BigDecimal asDecimalVersion(String version) {
...
     }
 
+    public boolean isMeasureEditable(Measure measure) {
+        featureFlagMap = featureFlag.findFeatureFlags();
+        return featureFlagMap.get(FeatureFlagConstant.FHIR_EDIT) || MeasureDetailsUtil.QDM.equals(measure.getMeasureModel());
+    }
+
+    public boolean isCqlLibraryEditable(CQLLibrary cqlLibrary) {
+        featureFlagMap = featureFlag.findFeatureFlags();
+        return featureFlagMap.get(FeatureFlagConstant.FHIR_EDIT) || MeasureDetailsUtil.QDM.equals(cqlLibrary.getLibraryModelType());
```
serhii-ilin: This server-side class has multiple changes not reflected in unit tests

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
RohitKandimalla: Done

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
RohitKandimalla: I removed it

src/main/java/mat/server/FeatureFlagServiceImpl.java
```diff
@@ -24,6 +25,7 @@
 	FeatureFlagDAO featureFlagDAO;
 
 	@Override
+	@Cacheable("featureFlags")
```
serhii-ilin: I think @adongare  told he needs it cached on dao level...

pom.xml
```diff
@@ -499,6 +499,11 @@
             <version>0.8.13</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
```
serhii-ilin: 1. why do you need this dependency. which code relies on that?
2. why do you specify that version on POM? our POM inherits dependency management from spring boot parent BOM so the correct version is already defined there

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
```diff
@@ -13,15 +14,27 @@
...
     private static BigDecimal QDM_VER_BEFORE_CONVERSION = new BigDecimal("5.5");
     private static BigDecimal MAT_VER_BEFORE_CONVERSION = new BigDecimal("5.8");
 
     private boolean isMeasure;
 
-    private static MatContextServiceUtil instance = new MatContextServiceUtil();
+    private static MatContextServiceUtil instance;
+
+    @Autowired
+    FeatureFlagService featureFlag;
```
serhii-ilin: btw it should be private

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
```diff
@@ -206,4 +228,11 @@ private BigDecimal asDecimalVersion(String version) {
         }
     }
 
+    public boolean isMeasureEditable(Measure measure) {
+        return featureFlag.findFeatureFlags().getOrDefault(FeatureFlagConstant.FHIR_EDIT, false) || MeasureDetailsUtil.QDM.equals(measure.getMeasureModel());
```
serhii-ilin: 1. I didn't notice that at first sight.. but.. it will make Pre-QDM measures editable too. Do we want that?
2. there are similar methods isEditable and isMeasureEditable which have different logic.. It can be confusing. How about we rename it into isMeasureModelEditable ?

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
```diff
@@ -206,4 +228,11 @@ private BigDecimal asDecimalVersion(String version) {
         }
     }
 
+    public boolean isMeasureEditable(Measure measure) {
+        return featureFlag.findFeatureFlags().getOrDefault(FeatureFlagConstant.FHIR_EDIT, false) || MeasureDetailsUtil.QDM.equals(measure.getMeasureModel());
+    }
+
+    public boolean isCqlLibraryEditable(CQLLibrary cqlLibrary) {
+        return featureFlag.findFeatureFlags().getOrDefault(FeatureFlagConstant.FHIR_EDIT, false) || MeasureDetailsUtil.QDM.equals(cqlLibrary.getLibraryModelType());
```
serhii-ilin: same as for measures

src/test/java/mat/server/service/impl/MatContextServiceUtilTest.java
```diff
@@ -55,4 +74,43 @@ public void testQdmVersionFail() {
...
+
+        measure.setMeasureModel("QDM");
+
+        assertEquals(true, matContextServiceUtil.isMeasureEditable(measure));
+    }
+
+    @Test
+    public  void testIsCqlLibraryEditableFail() {
+        featureFlagMap.put("FhirEdit", false);
+        Mockito.when(featureFlagService.findFeatureFlags()).thenReturn(featureFlagMap);
```
serhii-ilin: Pre-QDM tests are missing

src/main/java/mat/server/FeatureFlagServiceImpl.java
adongare: I want just one flag and not map or list of flags. if we cache this at dao layer, and if another dao needs it, I still need to perform this processing to find if flag is on/off. I would prefer a new method which I added already to find if flag is on/off by keeping caching at service level.

RohitKandimalla: @adongare As far as I remember, FindFeatureFlagByName(String name) implementation is replaced with Hashmap.

serhii-ilin: yeah.. actually DAO can be changed to return a MAP (I had originally proposed that in the code review before) and then it can be annotated as Cached.

pom.xml
RohitKandimalla: this dependency is for Spring cacheable

RohitKandimalla: I had build errors, so I had to use <version>5.2.2.RELEASE</version>

src/main/java/mat/server/FeatureFlagServiceImpl.java
adongare: okay, if we move map preparation to dao and add caching at dao layer, that will work.

src/main/java/mat/server/FeatureFlagServiceImpl.java
```diff
@@ -24,6 +25,7 @@
 	FeatureFlagDAO featureFlagDAO;
```
adongare: this one private as well

RohitKandimalla: Done

src/test/java/mat/server/service/impl/MatContextServiceUtilTest.java
RohitKandimalla: You mean model type as Pre-QDM ?

pom.xml
serhii-ilin: Strange, that should work out of the box. I need to check on my local

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
RohitKandimalla: I don't see any Pre-QDM measures.

serhii-ilin: You don't see it where? Ui has this logic, if model name is empty then it's pre cql/qdm

RohitKandimalla: yup, But server side doesn't know Pre-CQL. May be I should write test case for null values then.

pom.xml
serhii-ilin: <dependency>
    Did you try following?

<groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-starter-cache</artifactId>
</dependency>

src/test/java/mat/server/service/impl/MatContextServiceUtilTest.java
serhii-ilin: Yeah..

RohitKandimalla: ok

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
serhii-ilin: That what i meant

pom.xml
RohitKandimalla: No I will try Spring-boot.

src/main/java/mat/client/shared/CQLLibraryResultTable.java
RohitKandimalla: This is the requirement. "When the flag is off, the double click of a measure grid row does not launch the edit / view of a measure if the measure is a fhir measure but does for a QDM measure."

It doesn't change any existing logic

serhii-ilin: The requirements might be inconsistent.
Just make sure that 'View' functionality is not broken after that (you can see how it works on the measures you don't own in matdev).

RohitKandimalla: ooh ok, I will check it

pom.xml
RohitKandimalla: Yes spring-boot-starter-cache works fine

src/main/java/mat/client/shared/CQLLibraryResultTable.java
RohitKandimalla: View functionality is not broken from Grid button, but double of a measure which is viewable cannot be viewed. Since according to the requirement, edit / view of measure is disabled when double clicked.

serhii-ilin: Are you sure?

There was a bug fixed recently to make it work https://jira.cms.gov/browse/MAT-422

RohitKandimalla: That implies that, double click is not disabled in any case ?

serhii-ilin: that's my understanding. you can clarify with Kathy

RohitKandimalla: No Kathy says that, no matter what if the measure is FHIR, users cannot view or edit it.
From Kathy "If the fhiredit flag is off, we can't edit or view the fhir measure"

serhii-ilin: ok, great. thanks for the confirmation

src/main/java/mat/client/measure/ManageMeasureSearchModel.java
```diff
@@ -717,6 +722,15 @@ public int setClickCount(int val) {
         public void incrementClickCount() {
             this.clickCount++;
         }
+
+        public void setFhirEditOrViewable(boolean isFhirEditorViewable) {
+            this.isFhirEditorViewable = isFhirEditorViewable;
```
serhii-ilin: nice

war/WEB-INF/applicationContext-service.xml
```diff
@@ -24,9 +24,6 @@
 	<bean id="loginService" class="mat.server.service.impl.LoginCredentialServiceImpl">	    
 	</bean>
 	
-	<bean id="featureFlag" class="mat.server.FeatureFlagServiceImpl">	    
-	</bean>
-
```
serhii-ilin: Good. It was registered 2 times. 1st in xml and the other one with the annotation.
Same issue we have for the new gwt servlets

serhii-ilin:
The code change looks good to me

war/WEB-INF/applicationContext-service.xml
RohitKandimalla: Yup

PR #45, opened 2020-01-07
Import CQL Libraries - Measure Composer

adongare:
MAT-438 MAT - Import CQL Libraries - CQL Composer
MAT-438 MAT - Import CQL Libraries - CQL Composer
MAT-440 Update Help Text


src/main/java/mat/server/FeatureFlagServiceImpl.java
```diff
@@ -13,7 +13,7 @@
 import java.util.Map;
 import java.util.stream.Collectors;
 
-@Service
+@Service("featureFlagService")
```
serhii-ilin: why do we need a name? just curious. it's usually not required

src/main/java/mat/server/CQLServiceImpl.java
```diff
@@ -102,18 +103,21 @@
...
+
+	@Autowired
 	private MeasurePackageService measurePackageService;
-	
+
 	@Autowired
 	private UserDAO userDAO;
 
+	@Autowired
+	private FeatureFlagServiceImpl featureFlagService;
```
serhii-ilin: why do you inject am Impl and not a corresponding interface?

src/main/java/mat/server/FeatureFlagServiceImpl.java
adongare: we don't need this. was experimenting with something else. this service is registered twice with spring container. Need refactor.

src/main/java/mat/server/CQLServiceImpl.java
adongare: this is because the @Service is on impl and not on interface i guess. interface did not work

serhii-ilin: that should work in any case on interface. regardless of where the annotation is placed.
if it's not, then something is broken in our spring jar dependencies

serhii-ilin:
it has merge conflicts now

adongare:
> it has merge conflicts now

Resolved all the conflicts.

pom.xml
```diff
@@ -605,6 +605,10 @@
                     <argLine>
                         --illegal-access=permit
                     </argLine>
+                    <excludes>
+                        <!--run DAO tests manually -->
```
serhii-ilin: Probably we should come up with a better filter for the unit tests like that sometime later. But this is good for now.

adongare: Yup, Agreed

PR #50, opened 2020-01-13
MAT-182

RohitKandimalla:
XSS validation for Measure Name Field and Check Box when selecting a Composite Measure.

### DIFF
--- a/src/main/java/mat/client/measure/ComponentMeasureSearch.java
+++ b/src/main/java/mat/client/measure/ComponentMeasureSearch.java
@@ -303,7 +303,7 @@ public class ComponentMeasureSearch implements BaseDisplay{
 			public SafeHtml getValue(ManageMeasureSearchModel.Result object) {
 				SafeHtmlBuilder sb = new SafeHtmlBuilder();
 				sb.appendHtmlConstant("<div id='appliedComponentContainer' tabindex=\"-1\">");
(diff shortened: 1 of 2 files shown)


src/main/java/mat/client/measure/ComponentMeasureSearch.java
```diff
@@ -303,7 +303,8 @@ private void buildAppliedComponentMeasuresTableColumns() {
 			public SafeHtml getValue(ManageMeasureSearchModel.Result object) {
 				SafeHtmlBuilder sb = new SafeHtmlBuilder();
 				sb.appendHtmlConstant("<div id='appliedComponentContainer' tabindex=\"-1\">");
-				sb.appendHtmlConstant("<span id='appliedComponentDiv' title=\" " + object.getName() + "\" tabindex=\"0\">" + object.getName() + "</span>");
+				sb.appendHtmlConstant("<span id='appliedComponentDiv' title=\" " + SafeHtmlUtils.htmlEscape(object.getName()) +
```
adongare: Can you please check if the Ids `appliedComponentDiv`, `div2` etc are really needed.  If not you can remove them or use class.

RohitKandimalla: We have CSS properties associated with these ids. Should I change these styling using 'class'?

RohitKandimalla: Done !

PR #56, opened 2020-01-21
Feature/mat 496

RohitKandimalla:
Updated the Single and Double click events handled on CQL Grids, including checkbox selection.

### DIFF
--- a/src/main/java/mat/client/shared/CQLLibraryResultTable.java
+++ b/src/main/java/mat/client/shared/CQLLibraryResultTable.java
@@ -1,9 +1,7 @@
 package mat.client.shared;
 
-import com.google.gwt.cell.client.Cell;
 import com.google.gwt.cell.client.CheckboxCell;
-import com.google.gwt.dom.client.Element;
-import com.google.gwt.dom.client.NativeEvent;
+import com.google.gwt.dom.client.BrowserEvents;
 import com.google.gwt.event.logical.shared.HasSelectionHandlers;
 import com.google.gwt.event.logical.shared.SelectionEvent;
 import com.google.gwt.safehtml.shared.SafeHtml;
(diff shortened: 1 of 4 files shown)


src/main/java/mat/client/shared/CQLLibraryResultTable.java
```diff
@@ -107,27 +76,27 @@ public void onBrowserEvent(Cell.Context context, Element elem, CQLLibraryDataSet
...
             table.addColumn(model, SafeHtmlUtils.fromSafeConstant("<span title='Version'>" + "Model" + "</span>"));
 
+        table.addCellPreviewHandler(event -> {
+            String eventType = event.getNativeEvent().getType();
+            obj = event.getValue();
+            event.getNativeEvent().preventDefault();
+            if (BrowserEvents.CLICK.equalsIgnoreCase(eventType)) {
+                if(previousObj == null){
+                    clickEventsOnSelectedRows(event, selectionModel, obj, fireEvent);
+                } else if (obj.getId() == previousObj.getId()) {
```
serhii-ilin: Is it the same logic as for the Measures Gird?

RohitKandimalla: No, Here MultiSelectionModel isn't working as expected. It doesn't set the value as expected. Therefore, I have to handle the selected objects manually.

RohitKandimalla: for example, When we have MultiSelectionModel.setSelected as false. It immediately overrides it to true. and I couldn't see any other setSelected method being called other than what I intended to. Also selectionModel is a local variable, So I am not sure what is causing it to alter the value.

src/main/java/mat/client/shared/CQLLibraryResultTable.java
```diff
@@ -107,27 +76,27 @@ public void onBrowserEvent(Cell.Context context, Element elem, CQLLibraryDataSet
...
+                    clickEventsOnSelectedRows(event, selectionModel, obj, fireEvent);
+                } else {
+                    previousObj.setSelected(false);
+                    clickEventsOnSelectedRows(event, selectionModel, obj, fireEvent);
```
adongare: can pls check if this works

```
if(previousObj != null) {
      previousObj.setSelected(false);
}
clickEventsOnSelectedRows(event, selectionModel, obj, fireEvent);
```

RohitKandimalla: If previousObj is same as obj, then it isn't working. Since you are setting it false, before handling the event.

PR #57, opened 2020-01-22
Disabling Expression Builder Button for FHIR measures.

RohitKandimalla:


src/main/java/mat/client/cqlworkspace/definitions/CQLDefinitionsView.java
```diff
@@ -254,7 +257,11 @@ public void setWidgetReadOnly(boolean isEditable) {
 		getDefineButtonBar().getInsertButton().setEnabled(isEditable);
 		getDefineButtonBar().getTimingExpButton().setEnabled(isEditable);
 		getDefineButtonBar().getEraseButton().setEnabled(isEditable);
-		getDefineButtonBar().getExpressionBuilderButton().setEnabled(isEditable);
+		if (isSelectedObjectFhir()) {
```
serhii-ilin: does it make sense to use a one-liner instead of if-else block?

getDefineButtonBar().getExpressionBuilderButton().setEnabled(isEditable && !isSelectedObjectFhir.getAsBoolean());

serhii-ilin:
Is it possible to add a unit test?
You can try to create an instance of the view and pass a mock-closure and verify if it becomes editable  based in the parameters.

src/main/java/mat/client/cqlworkspace/definitions/CQLDefinitionsView.java
RohitKandimalla: Yup It makes sense. Will update it.

RohitKandimalla:
> Is it possible to add a unit test?
> You can try to create an instance of the view and pass a mock-closure and verify if it becomes editable based in the parameters.

There is no Unit test for this. I will try to create one.

PR #60, opened 2020-02-03
Feature/mat 478 insert fhir attributes

RohitKandimalla:
Inserting Fhir attributes and datatypes for FHIR measures and QDM attributes and Datatypes for QDM measure. Applies for both CQL measure workspace and CQL standalone workspace.


src/main/java/mat/client/cqlworkspace/AbstractCQLWorkspacePresenter.java
```diff
@@ -494,7 +494,7 @@ public void onSuccess(VsacApiResult result) {
 
 	protected void addCQLLibraryEditorViewHandlers() {
 		cqlWorkspaceView.getCQLLibraryEditorView().getExportErrorFile().addClickHandler(event -> exportErrorFile());
-		cqlWorkspaceView.getCQLLibraryEditorView().getInsertButton().addClickHandler(event -> buildInsertPopUp());
+		cqlWorkspaceView.getCQLLibraryEditorView().getInsertButton().addClickHandler(event -> buildInsertPopUp(MatContext.get().getCurrentMeasureModel()));
```
serhii-ilin: how is this compatible with standalone cql workspace implementation? I'm not sure if a standalone library can rely on a current measure model type.
should it use MatContext.get().getCurrentCQLLibraryModelType()) instead?

RohitKandimalla: addCQLLibraryEditorViewHandlers() is being used by both stand-alone and measure grid. It registers the event handlers when starting the application. It's an arbitrary situation.

src/main/java/mat/server/CQLConstantServiceImpl.java
```diff
@@ -10,31 +10,34 @@
...
 import mat.server.service.CodeListService;
 import mat.server.service.MeasureLibraryService;
 import mat.server.util.MATPropertiesService;
 import mat.server.util.QDMUtil;
 import mat.shared.cql.model.FunctionSignature;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.web.client.RestClientResponseException;
+import org.springframework.web.client.RestTemplate;
 
+@ContextConfiguration(classes = Application.class)
```
serhii-ilin: are you sure you need ContextConfiguration? first of all it resides in a test package and should not be available in tomcat

src/main/java/mat/server/CQLConstantServiceImpl.java
```diff
@@ -75,6 +78,12 @@
 
 	private static final long serialVersionUID = 1L;
 
+
+	@Value("${mat.fhirMatServices.url}")
+	private String fhirMatServicesUrl;
+
+	private final String fhirMatServiceResouceForAttributes = "/find";
```
serhii-ilin: JSL naming convention for final fields is upper case

src/main/java/mat/server/CQLConstantServiceImpl.java
```diff
@@ -83,7 +92,9 @@
 	
 	@Autowired
 	private MeasureLibraryService measureLibraryService;
-	
+
+	@Autowired
+	RestTemplate restTemplate;
```
serhii-ilin: private

src/main/java/mat/client/cqlworkspace/InsertAttributeBuilderDialogBox.java
```diff
@@ -75,15 +76,16 @@
 	private static final String VALUE_SETS = "Value Sets";
 	private static final String CODES = "Codes";
 	private static final String NULLABLE = "Nullable";
+	private static final String FHIR = "FHIR";
```
serhii-ilin: looks like it's not used...you use MeasureDetailsUtil.FHIR instead

RohitKandimalla: forgot to remove it.

src/main/java/mat/server/CQLConstantServiceImpl.java
RohitKandimalla: done

src/main/java/mat/server/CQLConstantServiceImpl.java
RohitKandimalla: Done

src/main/java/mat/server/CQLConstantServiceImpl.java
```diff
@@ -205,4 +220,28 @@ private CQLTypeContainer getCQLTypeInformation() {
 	private QDMContainer getQDMInformation() {
 		return QDMUtil.getQDMContainer();
 	}
+
+	public void getFhirAttributeAndDataTypes(CQLConstantContainer cqlConstantContainer) {
+		List<String> fhirAttributeList = new ArrayList<>();
+		List<String> fhirDataTypeList = new ArrayList<>();
+
+		try {
+			JSONArray fhirResponse = new JSONArray(restTemplate.getForObject(fhirMatServicesUrl + fhirMatServiceResouceForAttributes, String.class));
```
serhii-ilin: I would suggest to get rid of json library usage if possible.
Did you try spring template convenience methods for that? Like getForEntity (..... String[].class) ?
If it works, it will result in cleaner code.

src/main/java/mat/server/CQLConstantServiceImpl.java
RohitKandimalla: I am using it in retrieving external API hostname and port from MAT.properties.

src/main/java/mat/server/CQLConstantServiceImpl.java
serhii-ilin: it's not string[].class, we can use Pojo/Jackson to map an entry type

src/main/java/mat/client/cqlworkspace/AbstractCQLWorkspacePresenter.java
serhii-ilin: Right. But standalone CQL should rely on MatContext.get().getCurrentCQLLibraryModelType() instead of MatContext.get().getCurrentMeasureModel()

How about this.. 
Does it make sense to define 
abstract String AbstractCqlWorkspacePresenter::getCurrentModelType 

CQLMeasureWorkSpacePresenter::getCurrentModelType = () -> MatContext.get().getCurrentMeasureModel()
CQLStandaloneWorkSpacePresenter::getCurrentModelType = () -> MatContext.get().getCurrentCQLLibraryModelType()

Then, you can use getCurrentModelType in all 3 presenters

src/main/java/mat/server/CQLConstantServiceImpl.java
```diff
@@ -205,4 +220,28 @@ private CQLTypeContainer getCQLTypeInformation() {
...
+	public void getFhirAttributeAndDataTypes(CQLConstantContainer cqlConstantContainer) {
+		List<String> fhirAttributeList = new ArrayList<>();
+		List<String> fhirDataTypeList = new ArrayList<>();
+
+		try {
+			JSONArray fhirResponse = new JSONArray(restTemplate.getForObject(fhirMatServicesUrl + FHIR_MAT_SERVICES_RECOURSE_FOR_ATTRIBUTES, String.class));
+
+			for (int i = 0; i < fhirResponse.length(); i++) {
+
+				if(!fhirDataTypeList.contains(fhirResponse.getJSONObject(i).get("matDataTypeDescription").toString())) {
```
serhii-ilin: I would suggest to introduce a pojo with following fields: 
- matDataTypeDescription
- matDataTypeDescription
- matAttributeName

src/main/java/mat/server/CQLConstantServiceImpl.java
serhii-ilin: it's a pure test only annotation. should not be part of PROD classpass

src/main/java/mat/server/CQLConstantServiceImpl.java
RohitKandimalla: Working on it

src/test/java/mat/server/CQLConstantServiceImplTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.web.client.RestTemplate;
+
+import java.io.File;
+import java.net.URL;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+@ContextConfiguration(classes = DaoTestConfig.class)
```
serhii-ilin: I believe that ContextConfiguration and MocitoExtension annotations are not compatible.
Most likely 'ContextConfiguration' is not actually used. You just inject all mocks and unresolved fhirMatServicesUrl

src/main/java/mat/client/cqlworkspace/AbstractCQLWorkspacePresenter.java
RohitKandimalla: Implemented

src/main/java/mat/server/CQLConstantServiceImpl.java
RohitKandimalla: Created a separate service class

src/main/java/mat/server/CqlAttributesRemoteCall.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Autowired
+    private RestTemplate restTemplate;
+
+    @Value("${mat.fhirMatServices.url}")
+    private String fhirMatServicesUrl;
+
+    private final static String FHIR_MAT_SERVICES_RECOURSE_FOR_ATTRIBUTES = "/find";
+
+    List<String> fhirDataTypeList = new ArrayList<>();
```
serhii-ilin: Services should be stateless. It's basically a static shared field

RohitKandimalla: I shouldn’t be declaring it as Static?

src/main/java/mat/server/CqlAttributesRemoteCall.java
```diff
@@ -0,0 +1,52 @@
...
+        try {
+            fhirResponse = restTemplate.getForEntity(fhirMatServicesUrl + FHIR_MAT_SERVICES_RECOURSE_FOR_ATTRIBUTES, ConversionMapping[].class);
+        } catch (RestClientResponseException e) {
+            throw new MatRuntimeException(e);
+        }
+        ConversionMapping[] conversionMappings = fhirResponse.getBody();
+
+        for (ConversionMapping conversionMapping : conversionMappings) {
+
+            if(!fhirDataTypeList.contains(conversionMapping.getMatDataTypeDescription())) {
```
serhii-ilin: Basically you are using an array list to and check contains operation in a loop. It's potentially an O(N^2) operation. 
Should you use a Set collection (HashSet or LinkedHashSet if order is important) instead?

serhii-ilin: Also, if the data is not changed often we might want to cache attributed/data

pom.xml
```diff
@@ -506,6 +506,12 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-cache</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
```
serhii-ilin: nice!

src/main/java/mat/client/cqlworkspace/CQLMeasureWorkSpacePresenter.java
```diff
@@ -1904,6 +1906,11 @@ public void onSuccess(CQLQualityDataModelWrapper result) {
 		}
 	}
 
+	@Override
+	protected String getCurrentModelType(){
+		return MatContext.get().getCurrentMeasureModel();
```
serhii-ilin: nice!

src/main/java/mat/server/CqlAttributesRemoteCall.java
serhii-ilin: no! it's a bad practice to have any stateless components/services.

src/test/java/mat/server/CqlAttributesRemoteCallTest.java
```diff
@@ -0,0 +1,50 @@
...
+    @Mock
+    private RestTemplate restTemplate;
+
+    private CQLConstantContainer cqlConstantContainer = new CQLConstantContainer();
+
+    @InjectMocks
+    CqlAttributesRemoteCall cqlAttributesRemoteCall;
+
+    @Test
+    void TestGetFhirAttributeAndDataTypes() throws Exception {
```
serhii-ilin: naming conversion. It should start from a lower case char

src/main/java/mat/server/CqlAttributesRemoteCall.java
RohitKandimalla: The reason I used List, is because the existing functionality will not change for adding Fhir Attributes.

src/main/java/mat/server/CqlAttributesRemoteCall.java
RohitKandimalla: Changed it as a local object

src/main/java/mat/server/CqlAttributesRemoteCallService.java
```diff
@@ -37,16 +37,11 @@ public void getFhirAttributeAndDataTypes(CQLConstantContainer cqlConstantContain
...
-            }
-            if(!fhirAttributeList.contains(conversionMapping.getMatAttributeName())) {
-                fhirAttributeList.add(conversionMapping.getMatAttributeName());
-            }
+            fhirDataTypeSet.add(conversionMapping.getMatDataTypeDescription());
+            fhirAttributeSet.add(conversionMapping.getMatAttributeName());
         }
-        cqlConstantContainer.setFhirCqlDataTypeList(fhirDataTypeList);
-        cqlConstantContainer.setFhirCqlAttributeList(fhirAttributeList);
+        cqlConstantContainer.setFhirCqlDataTypeList(new ArrayList<>(fhirDataTypeSet));
```
serhii-ilin: makes sense

src/test/java/mat/server/CqlAttributesRemoteCallTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.io.File;
+import java.net.URL;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+@ExtendWith(MockitoExtension.class)
+class CqlAttributesRemoteCallTest {
+
+    @Value("https://localhost:9090")
```
serhii-ilin: Value annotation is not needed here. It's not used

src/main/resources/MAT.properties
```diff
@@ -36,6 +36,8 @@ mat.qdm.default.expansion.id=Most Recent Code System Versions in VSAC
 #At every 5th min
 mat.cache.expiry.time=300000
 
+mat.fhirMatServices.url=http://localhost:9090
```
serhii-ilin: probably should be passed as a system property.
we pass 
-DFHIR_ORCH_MEASURE_SRVC_URL=https://matdev.semanticbits.com/mat-fhir-services/orchestration/measure

serhii-ilin: And we will have to unify it soon..

src/test/java/mat/server/CqlAttributesRemoteCallTest.java
RohitKandimalla: Cleared

src/main/java/mat/server/ConversionMapping.java
```diff
@@ -0,0 +1,24 @@
+package mat.server;
+
+import lombok.*;
```
adongare: this can be avoided by updating intellij settings, Thanks to @serhii-ilin for suggestion!

PR #64, opened 2020-02-06
Mapping services URL configuration for DEV

RohitKandimalla:


src/main/java/mat/server/CqlAttributesRemoteCallService.java
```diff
@@ -15,7 +15,7 @@
     @Autowired
     private RestTemplate restTemplate;
 
-    @Value("${qdm.qicore.mapping.services.url}")
+    @Value("${QDM_QICORE_MAPPING_SERVICES_URL:http://localhost:9090}")
```
adongare: if this is a VM option, please update the README file where we have all the VM options documented

PR #69, opened 2020-02-12
MAT-418 fhir insert attributes - light box - left panel with mock data

ghost:
JIRA https://jira.cms.gov/browse/MAT-418
image
image


src/main/java/mat/client/cqlworkspace/shared/EditorDisplay.java
```diff
@@ -0,0 +1,27 @@
+package mat.client.cqlworkspace.shared;
+
+import com.google.gwt.user.client.TakesValue;
+import com.google.gwt.user.client.ui.HasText;
+import com.google.gwt.user.client.ui.RequiresResize;
+import edu.ycp.cs.dh.acegwt.client.ace.AceAnnotationType;
+import edu.ycp.cs.dh.acegwt.client.ace.AceEditorCursorPosition;
+
+public interface EditorDisplay extends RequiresResize, HasText, TakesValue<String> {
```
serhii-ilin: This interface is not directly used yet. I'm planning on gradually migrate all code from AceEditor class usage to that interface to decouple UI from the business logic in the Presenters.

src/main/java/mat/client/inapphelp/message/InAppHelpMessages.java
```diff
@@ -234,6 +234,22 @@
 			"<br>" +
 			"<b>Cancel Button:</b> the Cancel button will close the attribute builder and not save any changes or selections that were made inside of it.";
 
+	public static final String CQL_LIBRARY_ATTRIBUTE_MODAL_FHIR = "<b>Attributes By Data Type:</b> This field is optional and provides a list of the FHIR data types available. Making a choice in this field will not affect how the attribute looks once it is inserted into the CQL Editor; however, it will narrow down the list of attributes in the next field to only those attributes that are applicable to the data type chosen.<br>" +
+			"<br>" +
+			"<b>Attributes:</b> This field is a list of the attributes available. Attributes chosen must be compatible with the type of data in your expression. If a choice was made in the Attributes By Data Type field, this list will only show the attributes that are compatible with that data type.<br>" +
```
cday-semanticbits: This is picky and I realize there it is used all over MAT but you are not supposed to use &lt;b&gt; element anymore in html. You are supposed to either use strong or css font-weight. You get warned about it these days in most html syntax checkers. I will leave it up to you if you want to change it or not.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b

cday-semanticbits (approved):
Looks good

src/main/java/mat/client/inapphelp/message/InAppHelpMessages.java
serhii-ilin: Good to know! I will do my homework. Unfortunately, I'm not really a UI person.

In that particular case, it's just a placeholder for the text which is not prepared yet. It was copied from a similar QDM dialog and 'QDM' was replaced with 'FHIR' :)

src/main/java/mat/client/cqlworkspace/attributes/InsertFrhiAttributesDialogDisplay.java
```diff
@@ -0,0 +1,5 @@
+package mat.client.cqlworkspace.attributes;
+
+public interface InsertFrhiAttributesDialogDisplay {
```
adongare: typo I guess Insert**Frhi**AttributesDialogDisplay :)

adongare (approved):

src/main/java/mat/client/cqlworkspace/attributes/InsertFrhiAttributesDialogDisplay.java
serhii-ilin: Good catch! thank you!

PR #71, opened 2020-02-14
Implemented PackageV1 Feature Flag

RohitKandimalla:
Acceptance Criteria:

- The PackageV1 flag is off in Dev. When we begin the Fhir measure feature it will be on in Dev.
- When the flag is off, the Measure Packager tab is disabled when a Fhir Measure is selected. It is enabled for QDM measures.
- When the flag is on, the Measure Packager tab is enabled for Fhir or QDM Measures, per the package rules.

### DIFF
--- a/src/main/java/mat/client/MeasureComposerPresenter.java
+++ b/src/main/java/mat/client/MeasureComposerPresenter.java
@@ -3,6 +3,7 @@ package mat.client;
 import java.util.LinkedList;
 import java.util.List;
 
+import mat.client.event.MeasureEditEvent;
 import mat.client.util.FeatureFlagConstant;
 import org.gwtbootstrap3.client.ui.Button;
 
@@ -64,7 +65,6 @@ public class MeasureComposerPresenter implements MatPresenter, MeasureHeading, E
     private CQLMeasureWorkSpacePresenter cqlWorkspacePresenter;
     private static String MEASURE_COMPOSER = "MeasureComposer";
     private static String MEASURE_PACKAGER = "Measure Packager";
-    private static MeasureComposerPresenter instance = null;
 
 
     class EnterKeyDownHandler implements KeyDownHandler {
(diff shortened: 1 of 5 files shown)


src/main/java/mat/client/MeasureComposerPresenter.java
```diff
@@ -90,6 +93,13 @@ public static void setSubSkipEmbeddedLink(String name) {
         subSkipContentHolder.setFocus(true);
     }
 
+    public static MeasureComposerPresenter getInstance() {
```
serhii-ilin: it would be great if we can omit to create an explicit singleton...

src/main/java/mat/client/shared/MeasureLibraryResultTable.java
```diff
@@ -107,6 +108,7 @@ public void run() {
                 } else if (obj.getClickCount() == 2  && obj.isFhirEditOrViewable()) {
                     singleClickTimer.cancel();
                     obj.setClickCount(0);
+                    MeasureComposerPresenter.getInstance().disableMeasurePackageTabForFhirMeasures(obj.getMeasureModel());
```
serhii-ilin: It introduces a circular dependency... between the instances
can we try to utilize an event-based approach here?

You could try to subscribe MeasureComposerPresenter to MeasureSelectedEvent and MeasureEditEvent

RohitKandimalla: Implemented

RohitKandimalla:
> cool!
> does it work? :)

Yes It does.

src/main/java/mat/client/MeasureComposerPresenter.java
```diff
@@ -192,6 +195,20 @@ public void onContinueToMeasurePackage(ContinueToMeasurePackageEvent event) {
                 buttonBar.setPageNamesOnState();
             }
         });
+
+        MatContext.get().getEventBus().addHandler(MeasureEditEvent.TYPE, event -> disableMeasurePackageTabForFhirMeasures(MatContext.get().getCurrentMeasureModel()));
```
serhii-ilin: isn't it beautiful with events? ;)

PR #73, opened 2020-02-17
Implemented Story MAT-584 Allow user to clear selected grid rows.

cday-semanticbits:
https://jira.cms.gov/browse/MAT-584
- Added a label showing # of selected rows.
- Added a link that only shows up when a row is selected to clear selected rows.
- Added a missing env var in README.md


serhii-ilin:
nice!
is it possible to add some unit tests?

PR #75, opened 2020-02-18
MAT-621 Inserting FHIR attributes with LightBox - Connect left panel to Real Service

ghost:
MAT-621  LightBox Left Panel - connect to real service

image


cday-semanticbits (approved):
Looks good to me.

adongare (approved):

PR #79, opened 2020-02-20
MAT-522: Automated Password and User Activity Evaluation

jkotanchik-SB:
https://jira.cms.gov/browse/MAT-522

Updating existing scheduled job to flag passwords as expired when the password is 50+ days old and send a warning email at 35 days:
- Corrected path to password expiration email template in properties.
- Renamed email templates.
- Password age limit value set to -44. Negative as Apache Date Utils only has add method and 44 to send email 5 days ahead of required change. The job sets the temp password flag to true which tells the login service to force the user to change their password. 

Updating existing scheduled job to flag accounts as inactive when not signed in for 60+ days and send a warning email at 30 days:
- Renamed email templates.
- Account inactivity time limit set to -60. Negative to use Apache Date Utils add method. Warning email set to -30. 

Added unit tests

Updated email templates and code comments with new age values.


src/main/java/mat/server/service/jobs/CheckUserChangePasswordLimit.java
```diff
@@ -22,15 +21,9 @@
...
-import mat.dao.EmailAuditLogDAO;
-import mat.dao.UserDAO;
-import mat.model.EmailAuditLog;
-import mat.model.User;
-import mat.server.service.UserService;
-import mat.server.util.ServerConstants;
-import mat.shared.ConstantMessages;
+import java.io.IOException;
+import java.text.SimpleDateFormat;
+import java.util.*;
```
serhii-ilin: individual imports are preferred in MAT. could you update your IntelliJ settings?

src/test/java/mat/server/service/jobs/CheckUserChangePasswordLimitTest.java
```diff
@@ -1,40 +1,119 @@
...
 
-import static org.junit.Assert.assertNotNull;
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
 
 public class CheckUserChangePasswordLimitTest extends MatAppContextTest {
     @MockBean
     UserDAO userDAO;
-    @Autowired
+    @MockBean
```
serhii-ilin: hm, I didn't design MatAppContextTest for an extension. it was just a way to test the app context. should we introduce a parent base context tets class? :)

serhii-ilin:
btw: there are some codacy warnings

jkotanchik-SB:
Working through them now. They included the on demand import change you mentioned above.

PR #84, opened 2020-02-21
Added security banner after login. 

cday-semanticbits:
Added security banner after login. https://jira.cms.gov/browse/MAT-659

- Added a new modal class.
- Added the banner after login was verified as success in all login cases (normal, first time, etc).


serhii-ilin:
wow, that's a lot of code

PR #87, opened 2020-02-25
Feature/mat 452 fhir cql validations

RohitKandimalla:
- CQL Expression Validations for Fhir measures/libraries. Validations are performed by Cql_qlm_translation microservice. 
- Validations works not only for parameter tab but also for the cql changes from 'function' and 'definition' tabs.

Navigation:
Measure Library / Measure Composer / CQL Workspace / Parameters
CQL Composer / CQL Library Workspace / Parameters


src/main/java/mat/server/CqlValidatorRemoteCallService.java
```diff
@@ -0,0 +1,40 @@
...
+    @Autowired
+    private RestTemplate restTemplate;
+
+    @Value("${CQL_ELM_TRANSLATION_URL:http://localhost:7070}")
+    private String cqlElmtranslationUrl;
+
+    private String CQL_VALIDATION_RESOURCE = "/cql/translator/cql";
+
+    public String validateCqlExpression(String cqlFileString) {
+
```
serhii-ilin: Was it possible to map a POJO response object with Jason instead of String?

serhii-ilin: Other than that it looks OK to me.
there is a codacity waring

RohitKandimalla: I am looking into if I could map it to POJO.

PR #89, opened 2020-02-26
MAT-425 Implement FHIR conversion service with FHIR CQL

ghost:


src/main/java/mat/model/cql/CQLIncludeLibrary.java
```diff
@@ -7,217 +7,161 @@
...
+    public static class Comparator implements java.util.Comparator<CQLIncludeLibrary>, IsSerializable {
+
+        @Override
+        public int compare(CQLIncludeLibrary o1, CQLIncludeLibrary o2) {
+            return o1.getAliasName().compareTo(o2.getAliasName());
+        }
+
+    }
+
+    public String toString() {
```
cday-semanticbits: I would use lombok here for `@ToString` generation.

src/main/java/mat/server/CodeListServiceImpl.java
```diff
@@ -23,151 +23,80 @@
...
+            if ((dataSet.getOid() != null) && !dataSet.getOid().equals(ConstantMessages.GENDER_OID)
+                    && !dataSet.getOid().equals(ConstantMessages.RACE_OID) && !dataSet.getOid().equals(ConstantMessages.ETHNICITY_OID)
+                    && !dataSet.getOid().equals(ConstantMessages.PAYER_OID)) {
+                filteredQDSElements.add(dataSet);
+            } else {
+                System.out.println();
+            }
+
+        }
+        Collections.sort(filteredQDSElements, new Comparator<QualityDataSetDTO>() {
```
cday-semanticbits: I would use a lambda here. It is cleaner imo.

src/main/java/mat/server/util/XmlProcessor.java
```diff
@@ -42,1198 +42,1167 @@
...
+        if (excp instanceof ParserConfigurationException) {
+            LOG.info("Document Builder Object creation failed"
+                    + excp.getStackTrace());
+        } else if (excp instanceof SAXException) {
+            LOG.info("Xml parsing failed:" + excp.getStackTrace());
+        } else if (excp instanceof IOException) {
+            LOG.info("Conversion of String XML to InputSource failed"
+                    + excp.getStackTrace());
+        } else {
+            LOG.info("Generic Exception: " +
```
cday-semanticbits: instead of e.printStackTrace() its better to use LOG.error and pass in the exception. That way you control where the exception gets dumped in the logging configuration.

cday-semanticbits (approved):
Mentioned a few small things, but looks good overall.

src/main/java/mat/model/clause/ModelTypeHelper.java
```diff
@@ -0,0 +1,30 @@
+package mat.model.clause;
+
+public final class ModelTypeHelper {
```
jkotanchik-SB: Would there be any benefit to converting this class to an Enum?

serhii-ilin: I would gladly convert it into an enum. But there are some hibernate mapping concerns and that's a lot of changes on server-side and UI end. So, I'm a bit hesitant about that.

jkotanchik-SB: Fair enough, I was thinking there might be a large testing effort behind the idea.

adongare (approved):
LGTM

src/main/java/mat/server/CQLServiceImpl.java
```diff
@@ -414,8 +414,7 @@ public SaveUpdateCQLResult saveAndModifyFunctions(String xml, CQLFunctions funct
             result.getCqlModel().setCqlFunctions(sortFunctionList(result.getCqlModel().getCqlFunctions()));
             result.setSuccess(true);
         } catch (Exception e) {
-            e.printStackTrace();
-            result.setSuccess(false);
+            logger.error(e.getMessage(), e);
```
jkotanchik-SB: Should the `result.setSuccess(false);` be left in? I saw it was left in other methods.

serhii-ilin: Well, it's false by default. But, maybe I will leave it for clarify

serhii-ilin: good catch

PR #96, opened 2020-02-28
[MAT-522] Expire inactive accounts past limit

jkotanchik-SB:
Updated logic in user inactivity check to mark accounts past the limit inactive. 
- Previously, check would only mark accounts inactive if the age matched the limit exactly.
- Warning emails are still sent only when the inactivity age matches the warning limit.
- Added unit test to verify duplicate warning emails are not sent.


src/main/java/mat/server/service/jobs/CheckUserLastLoginTask.java
```diff
@@ -210,7 +210,8 @@ private void checkUserLoginDays(final long noOfDayLimit, final String emailType)
 
             lastSignInDate = DateUtils.truncate(lastSignInDate, Calendar.DATE);
             logger.info(USER_LOG_LABEL + user.getFirstName() + "  :::: lastSignInDate :::::   " + lastSignInDate);
-            if (lastSignInDate.equals(daysAgo)) {
+            if (emailType.equals(WARNING_EMAIL_FLAG) && lastSignInDate.equals(daysAgo)
```
cday-semanticbits: You could refactor this into a private method that takes an emailType dateToCheck emailType and daysAgo as params and remove the duplicate code here and in 201/202 would prob make the if statement easier to read as well.

jkotanchik-SB: Nice idea!

PR #98, opened 2020-03-03
[MAT-650] disable bonnie integration

jkotanchik-SB:
- Added FhirBonnie feature flag. Turned off for DEV and prod.
- "Package and Upload to Bonnie" will be disabled for FHIR measures when the FhirBonnie feature flag is off.
- "Upload to Bonnie" link in the Export page is hidden for FHIR measures when the FhirBonnie feature flag is off.

### DIFF
--- a/src/main/java/liquibase/changelog.xml
+++ b/src/main/java/liquibase/changelog.xml
@@ -13,6 +13,7 @@
     <include file="classpath:/liquibase/addConvertedFromMeasureColumn.xml"/>
     <include file="classpath:/liquibase/addMeasureDetailsReferenceType.xml"/>
     <include file="classpath:/liquibase/addFhirDTlFeatureFlags.xml"/>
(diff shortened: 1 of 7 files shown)


src/main/java/liquibase/changelog.xml
```diff
@@ -13,6 +13,7 @@
     <include file="classpath:/liquibase/addConvertedFromMeasureColumn.xml"/>
     <include file="classpath:/liquibase/addMeasureDetailsReferenceType.xml"/>
     <include file="classpath:/liquibase/addFhirDTlFeatureFlags.xml"/>
+    <include file="classpath:/liquibase/addFhirBonnieFeatureFlag.xml"/>
```
serhii-ilin: Maybe not critical in that particular case. But I would suggest appending new change-sets at the end of the list.

PR #100, opened 2020-03-05
[MAT-320 & MAT-703] ExportV1 feature flag impl & Adding MAT version to export button hover text

jkotanchik-SB:
Implemented checks for ExportV1 feature flag. When the flag is disabled, fhir measures cannot be exported.

When the ExportV1 flag is disabled:
- Library grid export button will be disabled when a fhir measure is selected first, but enabled when an exportable QDM measure is selected first.
- Bulk export will exclude fhir measures in addition to non-exportable QDM measures.


serhii-ilin:
Is the corresponding liquibase change in a separate PR?

jkotanchik-SB:
> Is the corresponding liquibase change in a separate PR?

It's already in develop in the addMVPFeatureFlags.xml script.

PR #106, opened 2020-03-06
Removed dictionary word check on passwords.

cday-semanticbits:
Removed dependencies used for dictionary check.
Removed dead code associated with dictionary check.


serhii-ilin:
We should use JIRA number in the branch name and in corresponding commits for tracking purposes. Otherwise, there is no way to link it to the requirements.

PR #111, opened 2020-03-06
Used Decimal Format, Refactored code.

RohitKandimalla:


RohitKandimalla:
From Jira - In the measure grids, display three decimal points for draft measures.  They are only currently displaying 1 decimal.

Unable to replicate the same in Local, but making sure that we display 3 digits in revision Number field for draft measures/Libraries.

PR #112, opened 2020-03-09
Limiting the creation of new composite measures to QDM

RohitKandimalla:
image

- This is not behind a feature flag. Composite FHIR measures are turned off until the standards are agreed upon.
- In the Create New Composite Measure, the Model radio button is set to QDM and is disabled.

### DIFF
--- a/src/main/java/mat/client/measure/AbstractNewMeasureView.java
+++ b/src/main/java/mat/client/measure/AbstractNewMeasureView.java
@@ -278,6 +278,16 @@ public class AbstractNewMeasureView implements DetailDisplay {
         return measureModelPanel;
     }
 
+    protected VerticalPanel buildCompositeModelTypePanel() {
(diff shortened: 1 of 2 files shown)


src/main/java/mat/client/measure/AbstractNewMeasureView.java
```diff
@@ -278,6 +278,16 @@ protected VerticalPanel buildModelTypePanel() {
         return measureModelPanel;
     }
 
+    protected VerticalPanel buildCompositeModelTypePanel() {
+        VerticalPanel measureModelPanel = new VerticalPanel();
```
adongare: I would avoid using Vertical/HorizontalPanels. They have some performance issues because they get translated to tables. I did the same mistake while implementing FHIR/QDM radio buttons. FlowPanel is better for this use case. we have VerticalFlowPanel(custom flow panel). Should work

PR #113, opened 2020-03-09
Fixed MAT-146 + Technical Debt cleanup

cday-semanticbits:
Fixed MAT-146. What was going on was there were two spring beans and getBean was being called which threw an exception preventing cql exports. 

To fix this i went ahead and did some technical debt cleanup. I removed all the xml configured beans and did it all with annotations. After doing this I also had to go in and fix the places where the bean names were referenced.

### DIFF
--- a/src/main/java/mat/client/export/measure/ManageMeasureExportView.java
+++ b/src/main/java/mat/client/export/measure/ManageMeasureExportView.java
@@ -1,5 +1,6 @@
 package mat.client.export.measure;
 
+import org.apache.commons.lang3.StringUtils;
 import org.gwtbootstrap3.client.ui.Button;
 import org.gwtbootstrap3.client.ui.ButtonToolBar;
 import org.gwtbootstrap3.client.ui.FormGroup;
@@ -118,7 +119,7 @@ public class ManageMeasureExportView implements ExportDisplay {
 		
 		vp.add(humanReadableRadio);
 		vp.add(hqmfRadio);
-		if(releaseVersion.startsWith("v5")){
+		if(isV5OrGreater(releaseVersion)){
 			vp.add(cqlLibraryRadio);
 			vp.add(elmRadio);
 			vp.add(jsonRadio);
@@ -127,6 +128,17 @@ public class ManageMeasureExportView implements ExportDisplay {
 		vp.add(compositeMeasurePackageRadio);
 		resetRadioButtonValues(isCompositeMeasure);
 	}
+
+	private boolean isV5OrGreater(String version) {
+		boolean result = false;
+		try {
+			result = Double.parseDouble(version.substring(1,version.length())) >= 5.0;
+		} catch (RuntimeException re) {
+		}
+		return result;
+	}
+
+
 	
 	@Override
 	public void showCompositeMeasure(boolean isComposite) {
--- a/src/main/java/mat/client/export/measure/ManageMeasureExportView.java
+++ b/src/main/java/mat/client/export/measure/ManageMeasureExportView.java
@@ -133,7 +133,7 @@ public class ManageMeasureExportView implements ExportDisplay {
 		boolean result = false;
 		try {
 			result = Double.parseDouble(version.substring(1,version.length())) >= 5.0;
-		} catch (RuntimeException re) {
+		} catch (NullPointerException|NumberFormatException re) {
 			//Just eat it.
 		}
 		return result;
--- a/src/main/java/mat/client/export/measure/ManageMeasureExportView.java
+++ b/src/main/java/mat/client/export/measure/ManageMeasureExportView.java
@@ -134,6 +134,7 @@ public class ManageMeasureExportView implements ExportDisplay {
 		try {
 			result = Double.parseDouble(version.substring(1,version.length())) >= 5.0;
 		} catch (RuntimeException re) {
+			//Just eat it.
 		}
 		return result;
 	}
--- a/src/main/java/mat/server/Application.java
+++ b/src/main/java/mat/server/Application.java
@@ -46,6 +46,7 @@ import org.springframework.security.core.userdetails.User;
 import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.security.core.userdetails.UserDetailsService;
 import org.springframework.security.provisioning.InMemoryUserDetailsManager;
+import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.EnableTransactionManagement;
 import org.springframework.web.client.RestTemplate;
 
@@ -56,7 +57,6 @@ import mat.server.twofactorauth.OTPValidatorInterfaceForUser;
 import mat.server.util.MATPropertiesService;
 
 @Configuration
-@ComponentScan({"mat.model", "mat.dao", "mat.dao.impl", "mat.model.clause", "mat.server", "mat.hibernate"})
 @PropertySource("classpath:MAT.properties")
 @EnableTransactionManagement
 @EnableConfigurationProperties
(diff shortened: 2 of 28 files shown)


src/main/java/mat/client/export/measure/ManageMeasureExportView.java
```diff
@@ -127,6 +128,17 @@ public void setExportOptionsBasedOnVersion(String releaseVersion, boolean isComp
 		vp.add(compositeMeasurePackageRadio);
 		resetRadioButtonValues(isCompositeMeasure);
 	}
+
+	private boolean isV5OrGreater(String version) {
+		boolean result = false;
+		try {
+			result = Double.parseDouble(version.substring(1,version.length())) >= 5.0;
+		} catch (RuntimeException re) {
+		}
```
serhii-ilin: Codacy will complain on a catch-and-swallow block. It should have a comment or a logging statement

src/main/java/mat/server/Application.java
```diff
@@ -56,14 +57,14 @@
...
 @Configuration
-@ComponentScan({"mat.model", "mat.dao", "mat.dao.impl", "mat.model.clause", "mat.server", "mat.hibernate"})
 @PropertySource("classpath:MAT.properties")
 @EnableTransactionManagement
 @EnableConfigurationProperties
 @EnableWebSecurity
 @EnableCaching
 @EnableScheduling
 @EnableJpaRepositories
+@Service
```
serhii-ilin: why do you need a service annotation on Application class? It's probably not needed

cday-semanticbits: It's a war remember, not a spring boot application. That just ensures it gets scanned.

src/main/java/mat/client/export/measure/ManageMeasureExportView.java
cday-semanticbits: ok will add a comment.

src/main/java/mat/client/export/measure/ManageMeasureExportView.java
```diff
@@ -127,6 +128,17 @@ public void setExportOptionsBasedOnVersion(String releaseVersion, boolean isComp
 		vp.add(compositeMeasurePackageRadio);
 		resetRadioButtonValues(isCompositeMeasure);
 	}
+
+	private boolean isV5OrGreater(String version) {
+		boolean result = false;
+		try {
+			result = Double.parseDouble(version.substring(1,version.length())) >= 5.0;
+		} catch (RuntimeException re) {
```
jkotanchik-SB: `RuntimeException` seems overly broad. Could it be some combination of other exceptions? If we get up to a runtime exception, there's probably something else we should be doing.

src/main/java/mat/server/clause/QDSAttributesServiceImpl.java
```diff
@@ -29,11 +29,13 @@
 import mat.model.clause.QDSAttributes;
 import mat.server.SpringRemoteServiceServlet;
 import mat.server.util.ResourceLoader;
+import org.springframework.stereotype.Service;
 
 /**
  * The Class QDSAttributesServiceImpl.
  */
 @SuppressWarnings("serial")
+@Service
```
serhii-ilin: It's a servlet. It should not have a Service annotation (if it's not injected somewhere else).

src/main/java/mat/client/export/measure/ManageMeasureExportView.java
cday-semanticbits: It handles all the cases I need NullPointer,NumberFormat, etc. Why create multiple blocks and increase the code size?

src/main/java/mat/server/clause/QDSAttributesServiceImpl.java
cday-semanticbits: Wat its not a servlet? its a spring bean.

src/main/java/mat/server/Application.java
serhii-ilin: I think it has "@Configuration" annotation which should be scanned by default.

serhii-ilin:
overall looks good to me.  a couple of minor remarks and complains from codacy

src/main/java/mat/client/export/measure/ManageMeasureExportView.java
serhii-ilin: It's java 11.  catch(NullPointer|IllegalArgumentException e) { // ... }
Indeed. Most of code scanning toold will complain about a borad exception :)

src/main/java/mat/server/Application.java
jkotanchik-SB: From [here](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Configuration.html):

> @Configuration is meta-annotated with @Component, therefore @Configuration classes are candidates for component scanning

src/main/java/mat/server/service/impl/MeasurePackageServiceImpl.java
```diff
@@ -122,7 +125,8 @@
     @Autowired
     private CompositeMeasurePackageValidator compositeMeasurePackageValidator;
 
-    private String currentReleaseVersion;
+    @Value("${mat.measure.current.release.version}")
+    private String currentReleaseVersion = "v6.0";
```
jkotanchik-SB: What happens when you assign a value and use the value annotation?

src/main/java/mat/server/service/impl/SimpleEMeasureServiceImpl.java
```diff
@@ -246,7 +246,7 @@ private void findAndAddDTO(final NodeList nodeList, final List<QualityDataSetDTO
 	 * @return Measure. *
 	 */
 	private mat.model.clause.Measure getMeasureName(final String measureId) {
-		MeasurePackageService measureService = (MeasurePackageService) context.getBean("measurePackageService");
+		MeasurePackageService measureService = (MeasurePackageService) context.getBean(MeasurePackageService.class);
```
jkotanchik-SB: Can remove the type cast.

src/main/java/mat/server/service/impl/MeasurePackageServiceImpl.java
cday-semanticbits: it overrides what was there from the property. need to remove the string.

src/main/java/mat/server/service/impl/SimpleEMeasureServiceImpl.java
cday-semanticbits: removing

src/main/java/mat/server/clause/QDSAttributesServiceImpl.java
cday-semanticbits: See what you mean now, my bad. It was a bean before but prob someone new to spring. Going to try removing it and see if it works.

src/main/java/mat/client/export/measure/ManageMeasureExportView.java
cday-semanticbits: nice, switched over to java11 style.

src/main/java/mat/server/clause/QDSAttributesServiceImpl.java
cday-semanticbits: It worked no need for it to be a bean.

src/main/java/mat/client/export/measure/ManageMeasureExportView.java
serhii-ilin: sorry for misleading you.
it's a Java7 feature
https://docs.oracle.com/javase/7/docs/technotes/guides/language/catch-multiple.html

PR #115, opened 2020-03-10
New Export Page for Fhir measures

RohitKandimalla:


lib/mat/mat-rest-commons/0.0.1-SNAPSHOT/maven-metadata-local.xml
```diff
@@ -7,17 +7,17 @@
     <snapshot>
       <localCopy>true</localCopy>
     </snapshot>
-    <lastUpdated>20200305190420</lastUpdated>
```
cday-semanticbits: we should .gitignore these files in lib/mat IMO. We shouldn't be checking these in every time.

RohitKandimalla:
> No, you don't want to ignore them. we should not just update/commit them.
> I moved install plugin into a dedicated profile so that it doesn't happen often.

ok Will revert it back

src/main/java/mat/client/export/measure/ManageMeasureExportView.java
```diff
@@ -223,4 +238,14 @@ public boolean isCompositeMeasurePackageRadio() {
...
 	}
+
+	@Override
+	public boolean isXml() {
+		return xmlRadio.getValue();
+	}
+
+	@Override
+	public boolean isAll() {
+		return allRadio.getValue();
```
serhii-ilin: Looks like you use 'TABS' instead of 'SPACES' for formatting your code

lib/mat/mat-rest-commons/0.0.1-SNAPSHOT/maven-metadata-local.xml
serhii-ilin: @cday-semanticbits - that will affect how the maven repo works. It's a workaround for the dependencies. But there is another PR where I moved 'install-plugin' to a dedicated profile, so we don't update the repo files every time.

src/main/java/mat/client/export/measure/ManageMeasureExportView.java
RohitKandimalla: image
It looks good to me in local.

RohitKandimalla:
> Looks good
> But let's not push maven's meta data files.

So you want me to revert back metadata files and push only java files?

serhii-ilin:
@RohitKandimalla 
`So you want me to revert back metadata files and push only java files?`
Exactly. If you haven't changed the libraries, there is no need to updated their meta-data..

PR #116, opened 2020-03-10
MAT-509 Add new FHIR measure in MAT

ghost:
MAT-509. MAT - Add New FHIR Measure

See https://jira.cms.gov/browse/MAT-509 for details.
FHIR resource is not needed at that point.

### DIFF
--- a/src/main/java/mat/server/util/MeasureUtility.java
+++ b/src/main/java/mat/server/util/MeasureUtility.java
@@ -103,10 +103,7 @@ public class MeasureUtility {
      * @param version
      */
     public static void updateCQLVersion(XmlProcessor processor, String version) throws XPathExpressionException {
-        Node versionNode = processor.findNode(processor.getOriginalDoc(), "//cqlLookUp/version");
-        if (versionNode != null) {
-            versionNode.setTextContent(version);
-        }
+        findAndReplace(processor, version, "//cqlLookUp/version");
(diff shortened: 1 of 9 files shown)


RohitKandimalla (approved):

src/main/java/mat/server/util/MeasureUtility.java
```diff
@@ -98,24 +96,38 @@ public static String cleanString(String originalString) {
...
+     * Update CQL version.
+     *
+     * @param processor the processor
+     * @param version
+     */
+    public static void updateCQLVersion(XmlProcessor processor, String version) throws XPathExpressionException {
+        Node versionNode = processor.findNode(processor.getOriginalDoc(), "//cqlLookUp/version");
+        if (versionNode != null) {
+            versionNode.setTextContent(version);
+        }
```
jkotanchik-SB: Just a suggestion. Could refactor the search and replace logic into its own method that takes processor, xpath, and replacement text to reduce code duplication.

jkotanchik-SB (approved):

src/main/java/mat/server/util/MeasureUtility.java
serhii-ilin: Agree, makes sene. I could not do that due to error handling differences. But now we can do that.

PR #119, opened 2020-03-11
Fixed some gui issues when conversion/validation returns nulls in fields.

cday-semanticbits:
Added null protection on fields so the report will still show if they are null.

### DIFF
--- a/src/main/java/mat/server/ExportServlet.java
+++ b/src/main/java/mat/server/ExportServlet.java
@@ -1,7 +1,6 @@
 package mat.server;
 
 import java.io.IOException;
-import java.lang.invoke.MethodHandles;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.Collections;
@@ -55,9 +54,6 @@ import mat.shared.bonnie.error.BonnieServerException;
 import mat.shared.bonnie.error.BonnieUnauthorizedException;
 
 public class ExportServlet extends HttpServlet {
-    private static final Log logger = LogFactory.getLog(MethodHandles.lookup().lookupClass());
-
-
     private static final String LIBRARY_ID = "libraryid";
     private static final String USER_ID = "userId";
     private static final String EXPORT_MEASURE_OWNER = "exportMeasureOwner";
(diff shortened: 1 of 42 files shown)


src/main/java/mat/server/service/FhirValidationReportService.java
```diff
@@ -169,9 +174,15 @@ private String generateValidationReport(ConversionResultDto conversionResultDto,
 
             }
 
-            Instant instant = Instant.parse(conversionResultDto.getModified());
-            paramsMap.put("runDate", DateUtility.formatInstant(instant, DATE_FORMAT));
-            paramsMap.put("runTime", DateUtility.formatInstant(instant, TIME_FORMAT));
+            if (conversionResultDto.getModified() != null) {
+                Instant instant = Instant.parse(conversionResultDto.getModified());
```
serhii-ilin: JFYI. You could do that from the template with a default value.

src/main/java/mat/server/ExportServlet.java
```diff
@@ -554,6 +561,18 @@ private void exportSimpleXML(HttpServletResponse resp, String id, String type, M
         }
     }
 
+    private void exportMatXML(HttpServletResponse resp, String id, String type, Measure measure) throws Exception {
+        ExportResult export = getService().getMatXml(id);
+        if (SAVE.equals(type)) {
```
serhii-ilin: Is this part of a different JIRA? How is this related to the conversion?

src/main/java/mat/server/service/FhirValidationReportService.java
```diff
@@ -104,14 +105,15 @@ private String generateValidationReport(ConversionResultDto conversionResultDto,
...
+                if(StringUtils.isNotBlank(conversionResultDto.getErrorReason())) {
                     paramsMap.put("ErrorReason", conversionResultDto.getErrorReason());
                 }
                 // ValueSet FHIR validation errors
                 List<FhirValidationResult> valueSetFhirValidationErrors = new ArrayList<>();
                 if (CollectionUtils.isNotEmpty(conversionResultDto.getValueSetConversionResults())) {
                     for (ValueSetConversionResults valueSetConversionResults : conversionResultDto.getValueSetConversionResults()) {
-                        if (Boolean.FALSE.equals(valueSetConversionResults.getSuccess())) {
+                        if (valueSetConversionResults.getSuccess() != null &&
+                                false == valueSetConversionResults.getSuccess()) {
```
jkotanchik-SB: Why can't java just figured this out for itself.

PR #120, opened 2020-03-11
Fixed gui issues when conversion/validation returns null fields so the report still shows.

cday-semanticbits:


src/main/java/mat/server/service/FhirValidationReportService.java
```diff
@@ -166,12 +169,13 @@ private String generateValidationReport(ConversionResultDto conversionResultDto,
...
                 paramsMap.put("fhirMatCqlConversionErrors", fhirMatCqlConversionErrors);
-
             }
 
-            Instant instant = Instant.parse(conversionResultDto.getModified());
-            paramsMap.put("runDate", DateUtility.formatInstant(instant, DATE_FORMAT));
-            paramsMap.put("runTime", DateUtility.formatInstant(instant, TIME_FORMAT));
+            if (conversionResultDto.getModified() != null) {
+                Instant instant = Instant.parse(conversionResultDto.getModified());
+                paramsMap.put("runDate", DateUtility.formatInstant(instant, DATE_FORMAT));
```
serhii-ilin: Does this work?
I think you will have to modify validation_report_header.ftl

${runDate} -> ${runDate!""}
${runTime} -> ${runTime!""}

src/main/java/mat/server/service/FhirValidationReportService.java
```diff
@@ -129,7 +130,7 @@ private String generateValidationReport(ConversionResultDto conversionResultDto,
                         if("Not Found in Hapi".equals(results.getReason())) {
                             paramsMap.put("LibraryNotFoundInHapi", results.getReason());
                         }
-                        if (Boolean.FALSE.equals(results.getSuccess())) {
+                        if (results.getSuccess() == null || !results.getSuccess()) {
```
RohitKandimalla: According to Duane if results.getSuccues() is False, only then results.getLibraryFhirValidationResults() should be returning Errors. I am not sure if we should be considerring null.

src/main/java/mat/server/service/FhirValidationReportService.java
```diff
@@ -155,7 +156,9 @@ private String generateValidationReport(ConversionResultDto conversionResultDto,
 
                 //Measure FHIR validation errors
                 List<FhirValidationResult> measureFhirValidationErrors = new ArrayList<>();
```
jkotanchik-SB: Consider lazy instantiating the collections holding the reporting values throughout. (GH won't let me comment on code outside of the changed areas.)

Now I'm wondering, does lazy init do anything after this gets run through GWT?

src/main/java/mat/server/service/FhirValidationReportService.java
```diff
@@ -17,6 +17,7 @@
 import gov.cms.mat.fhir.rest.dto.LibraryConversionResults;
 import gov.cms.mat.fhir.rest.dto.MatCqlConversionException;
 import gov.cms.mat.fhir.rest.dto.ValueSetConversionResults;
+import liquibase.util.StringUtils;
```
serhii-ilin: That is the wrong StringUtils !!

src/main/java/mat/server/service/FhirValidationReportService.java
cday-semanticbits: these are fixes so it can be null and still function. that was one of the issues we faced.

src/main/java/mat/server/service/FhirValidationReportService.java
serhii-ilin: @jkotanchik-SB - this is server-side code. Not gwt

jkotanchik-SB: In that case, lazy init collections, when applied throughout, would reduce our footprint, though not by a lot.

PR #121, opened 2020-03-12
Fixed log statement to log the stack trace when there are conversion errors.

cday-semanticbits:
Fixed log statement to log the stack trace when there are conversion errors.


src/main/java/mat/server/clause/FhirMeasureRemoteServiceImpl.java
```diff
@@ -37,12 +37,9 @@ public FhirConvertResultResponse convert(Result sourceMeasure) throws MatExcepti
         }
         try {
             return fhirMeasureService.convert(sourceMeasure, vsacTicketInformation.getTicket(), LoggedInUserUtil.getLoggedInUser());
-        } catch (MatException e) {
-            logger.error(e);
+        } catch (RuntimeException|MatException e) {
```
serhii-ilin: Hm.. interesting..
Curious why the exception wrapping causes the issue..

cday-semanticbits: its logging the stack trace that is the issue here. my fix does that.

serhii-ilin:
lgtm

PR #122, opened 2020-03-12
Feature/hotfix merge

AndrewBird81:
The infamous "Tuple Bug" hotfix was never merged back into master. This corrects that.


serhii-ilin:
not sure why we need the poms changes

PR #125, opened 2020-03-13
MAT-738 Enable Run FHIR Validation button for Draft FHIR Measures

ghost:


jkotanchik-SB (approved):

src/main/java/mat/shared/model/util/MeasureDetailsUtil.java
```diff
@@ -112,16 +112,15 @@ public static String getModelTypeDisplayName(String type) {
...
+    private static boolean canValidateQdmMeasureVersions(Measure measure) {
         BigDecimal matVersion = asDecimalVersion(measure.getReleaseVersion());
         BigDecimal qdmVersion = asDecimalVersion(measure.getQdmVersion());
-
-        return (measure.isDraft() && measure.getMeasureModel().equals(FHIR))
-                || (!measure.isDraft() && measure.getMeasureModel().equals(QDM)
-                && RUN_FHIR_VALIDATION_VERSION.compareTo(matVersion) <= 0
-                && RUN_FHIR_VALIDATION_QDM_VERSION.compareTo(qdmVersion) <= 0);
+        return RUN_FHIR_VALIDATION_VERSION.compareTo(matVersion) <= 0
+                && RUN_FHIR_VALIDATION_QDM_VERSION.compareTo(qdmVersion) <= 0;
```
RohitKandimalla: Is it going to return the same boolean value as the previous code does? I mean do we have any logic change ?

jkotanchik-SB: Possibly different. if either `measure.getReleaseVersion()` or `measure.getQdmVersion()` is null and the other is the desired version, this will now return true. The null check at the beginning of `isValidatable` has been removed so the null would fall through to `aDecimalVersion()` which would return 0 and the compareTo would resolve to true. This would also happen if both version values are null.

RohitKandimalla (approved):

PR #126, opened 2020-03-13
fixing the bug in Mat Dev.

RohitKandimalla:


src/main/java/mat/server/util/MeasureUtility.java
```diff
@@ -69,6 +69,15 @@ public static String formatVersionText(String revisionNumber, String version) {
         return formatVersionText(version) + "." + revisionFormat.format(Integer.parseInt(revisionNumber));
     }
 
+    public static String formatVersionTextForIncludeLibraries(String version) {
+        StringUtility su = new StringUtility();
+        String[] versionArr = version.split("\\.");
```
serhii-ilin: should we handle safely cases when a version value has less than 3 or even 0 elements?

RohitKandimalla: Yes, working on it.

src/main/java/mat/server/util/MeasureUtility.java
```diff
@@ -69,6 +69,22 @@ public static String formatVersionText(String revisionNumber, String version) {
         return formatVersionText(version) + "." + revisionFormat.format(Integer.parseInt(revisionNumber));
     }
 
+    public static String formatVersionTextForIncludeLibraries(String version) {
+        StringUtility su = new StringUtility();
+        String[] versionArr = version.split("\\.");
+        String majorVersion = "", minorVersion = "", revisionNumber = "";
```
serhii-ilin: I believe most of the code style checkers recommend to define each variable on a separate line.

src/main/java/mat/server/util/MeasureUtility.java
```diff
@@ -69,6 +69,22 @@ public static String formatVersionText(String revisionNumber, String version) {
...
+        for(int i = 0; i < versionArr.length; i++) {
+            if(i == 0) {
+                majorVersion = su.trimLeadingZeros(versionArr[0]);
+            } else if(i == 1) {
+                minorVersion = su.trimLeadingZeros(versionArr[1]);
+            } else {
+                revisionNumber = su.trimLeadingZeros(versionArr[2]);
+            }
+        }
+        return majorVersion + "." + minorVersion + "." + revisionFormat.format(Integer.parseInt(revisionNumber == "" ? "000" : revisionNumber));
```
serhii-ilin: so, there is a possibility to see "..." version, interesting..

src/main/java/mat/server/util/MeasureUtility.java
RohitKandimalla: Updated it

PR #127, opened 2020-03-16
MAT-744 Fixed 500 measure conversion and validation issues.

cday-semanticbits:
MAT-744 Fixed 500 measure conversion and validation issues.
When cloning and packaging a new measure two title elements show up in the simple xml.
This fixes that.
It turns out one was intended to be used for measureModel but accidentally called title when added in back in October.

**IMPORTANT**
Also we need the new client jar from micro-services so that is in this story as well.


src/main/resources/MeasureDetailsModelMapping.xml
```diff
@@ -14,7 +14,7 @@
             <bind-xml name="title" node="element"/>
         </field>
         <field name="measureModel" type="java.lang.String">
-            <bind-xml name="title" node="element"/>
+            <bind-xml name="measureModel" node="element"/>
```
serhii-ilin: Nice

serhii-ilin:
does it affect only new measures but not the existing measures?
or we have legacy measures with this issue

cday-semanticbits:
@serhii-ilin It is not in master yet so we are good. Just existing cloned dev measures.

PR #128, opened 2020-03-16
MAT - 731 error converting

ghost:
https://jira.cms.gov/browse/MAT-731

MATDEV: Conversion Error 'An error occurred while converting measure. Please try again later.' Clicking X is has different behavior from Clicking 'Return To Measure Library'

Fixed
1. UI behavior difference
2. Use the same transaction for clone and save
3. Remove UMLS error when UMLS becomes active

### DIFF
--- a/src/main/java/mat/client/umls/ManageUmlsPresenter.java
+++ b/src/main/java/mat/client/umls/ManageUmlsPresenter.java
@@ -1,5 +1,8 @@
 package mat.client.umls;
 
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
 import org.gwtbootstrap3.client.ui.Button;
 import org.gwtbootstrap3.client.ui.FormGroup;
 import org.gwtbootstrap3.client.ui.HelpBlock;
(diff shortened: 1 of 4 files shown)


cday-semanticbits (approved):
lgtm

src/main/java/mat/client/umls/ManageUmlsPresenter.java
```diff
@@ -19,253 +19,261 @@
...
+        } else {
+            display.getUserIdGroup().setValidationState(ValidationState.SUCCESS);
+            display.getPasswordGroup().setValidationState(ValidationState.SUCCESS);
+            vsacapiService.validateVsacUser(display.getUserIdText().getText(),
+                    display.getPasswordInput().getText(), new AsyncCallback<Boolean>() {
+                        @Override
+                        public void onFailure(final Throwable caught) {
+                            display.getErrorMessageDisplay().setMessage(
+                                    MatContext.get().getMessageDelegate().getUML_LOGIN_UNAVAILABLE());
+                            caught.printStackTrace();
```
adongare: do we need to log this `printStackTrace()` ?

adongare (approved):

AndrewBird81 (approved):

src/main/java/mat/client/umls/ManageUmlsPresenter.java
serhii-ilin: well, i can switch to logging instead. I just re-formatted that code

PR #129, opened 2020-03-16
Couldn't find includes library because of version mismatch.

RohitKandimalla:
Solved the actual bug raised by Chris, But because of mismatch in formatting, MAT is unable to identify actual Includes libraries with similar versions. Therefore making the formatting unique.

### DIFF
--- a/src/main/java/mat/server/util/MeasureUtility.java
+++ b/src/main/java/mat/server/util/MeasureUtility.java
@@ -40,8 +40,19 @@ public class MeasureUtility {
     public static String formatVersionText(String version) {
         StringUtility su = new StringUtility();
         String[] versionArr = version.split("\\.");
-        String majorVersion = su.trimLeadingZeros(versionArr[0]);
-        String minorVersion = su.trimLeadingZeros(versionArr[1]);
+        String majorVersion = "";
+        String minorVersion = "";
+        String revisionNumber = "";
+        for(int i = 0; i < versionArr.length; i++) {
+            if(i == 0) {
+                majorVersion = su.trimLeadingZeros(versionArr[0]);
+            } else if(i == 1) {
+                minorVersion = su.trimLeadingZeros(versionArr[1]);
+            } else {
+                revisionNumber = su.trimLeadingZeros(versionArr[2]);
+                return majorVersion + "." + minorVersion + "." + revisionFormat.format(Integer.parseInt(revisionNumber));
+            }
+        }
         return majorVersion + "." + minorVersion;
     }
 
@@ -67,24 +78,6 @@ public class MeasureUtility {
      */
     public static String formatVersionText(String revisionNumber, String version) {
         return formatVersionText(version) + "." + revisionFormat.format(Integer.parseInt(revisionNumber));
-    }
-
-    public static String formatVersionTextForIncludeLibraries(String version) {
-        StringUtility su = new StringUtility();
-        String[] versionArr = version.split("\\.");
-        String majorVersion = "";
-        String minorVersion = "";
-        String revisionNumber = "";
-        for(int i = 0; i < versionArr.length; i++) {
-            if(i == 0) {
-                majorVersion = su.trimLeadingZeros(versionArr[0]);
-            } else if(i == 1) {
-                minorVersion = su.trimLeadingZeros(versionArr[1]);
-            } else {
-                revisionNumber = su.trimLeadingZeros(versionArr[2]);
-            }
-        }
-        return majorVersion + "." + minorVersion + "." + revisionFormat.format(Integer.parseInt("".equals(revisionNumber) ? "000" : revisionNumber));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/mat/server/util/MeasureUtility.java
```diff
@@ -40,8 +40,19 @@ public static String getVersionText(String version, boolean isDraft) {
...
+        String revisionNumber = "";
+        for(int i = 0; i < versionArr.length; i++) {
+            if(i == 0) {
+                majorVersion = su.trimLeadingZeros(versionArr[0]);
+            } else if(i == 1) {
+                minorVersion = su.trimLeadingZeros(versionArr[1]);
+            } else {
+                revisionNumber = su.trimLeadingZeros(versionArr[2]);
+                return majorVersion + "." + minorVersion + "." + revisionFormat.format(Integer.parseInt(revisionNumber));
+            }
```
jkotanchik-SB: I think this block could be condensed; looks like there are only 2 cases:
```
String majorVersion = su.trimLeadingZeros(versionArr[0]);
String minorVersion = su.trimLeadingZeros(versionArr[1]);
if (versionArr.length > 2) 
    return majorVersion + "." + minorVersion + "." + revisionFormat.format(Integer.parseInt(su.trimLeadingZeros(versionArr[2]));

return majorVersion + "." + su.trimLeadingZeros(versionArr[1]);
```

For revisionFormat, is there a benefit to converting the String to an int then formatting back to a String?

src/main/java/mat/server/util/MeasureUtility.java
```diff
@@ -40,8 +40,19 @@ public static String getVersionText(String version, boolean isDraft) {
...
+        String minorVersion = "";
+        String revisionNumber = "";
+        for(int i = 0; i < versionArr.length; i++) {
+            if(i == 0) {
+                majorVersion = su.trimLeadingZeros(versionArr[0]);
+            } else if(i == 1) {
+                minorVersion = su.trimLeadingZeros(versionArr[1]);
+            } else {
+                revisionNumber = su.trimLeadingZeros(versionArr[2]);
+                return majorVersion + "." + minorVersion + "." + revisionFormat.format(Integer.parseInt(revisionNumber));
```
jkotanchik-SB: For revisionFormat, is there a benefit to converting the String to an int then formatting back to a String?

src/test/java/mat/server/util/MeasureUtilityTest.java
```diff
@@ -17,7 +17,8 @@ public void testVersionTextIfDraftable() {
         assertEquals(MeasureUtility.getVersionText(version, isDraft), "Draft based on v4.2");
         assertEquals(MeasureUtility.getVersionTextWithRevisionNumber(version, revisionNumber, isDraft), "Draft v4.2.005");
         assertEquals(MeasureUtility.formatVersionText(revisionNumber, version), "4.2.005");
-        assertEquals(MeasureUtility.formatVersionTextForIncludeLibraries(version + "." + revisionNumber), "4.2.005");
+        assertEquals(MeasureUtility.formatVersionText(version + "." + revisionNumber), "4.2.005");
+        assertEquals(MeasureUtility.formatVersionText(version), "4.2");
```
jkotanchik-SB: JUnit asserts follow a pattern of `assert(expected, actual)`. Looks like the arguments are reversed in these asserts.

src/main/java/mat/server/util/MeasureUtility.java
RohitKandimalla: It is to make sure that at versionArr[0] / [1] it doesn't give out of bound exception.

src/main/java/mat/server/util/MeasureUtility.java
RohitKandimalla: Revision Number is the one that is causing an issue in MAT dev, not exactly sure why though. This is just a fix but was unable to find the root cause.

src/test/java/mat/server/util/MeasureUtilityTest.java
RohitKandimalla: my bad will update them

src/main/java/mat/server/util/MeasureUtility.java
jkotanchik-SB: Ah, are we assuming the incoming version is not well formed?

Does the calling method know how to handle a single dot, "3.", or "1..3"?

RohitKandimalla: Nope, it will display it as "1..3".
Ideally, the exiting code is exactly the same as you mentioned, which is 
```
String majorVersion = su.trimLeadingZeros(versionArr[0]);
String minorVersion = su.trimLeadingZeros(versionArr[1]);
if (versionArr.length > 2) 
    return majorVersion + "." + minorVersion + "." + revisionFormat.format(Integer.parseInt(su.trimLeadingZeros(versionArr[2]));

return majorVersion + "." + su.trimLeadingZeros(versionArr[1]);
```
I updated based on last CR comments to handle exception.

src/main/java/mat/server/util/MeasureUtility.java
jkotanchik-SB: Nevermind, I see now that it is converting the string to an int to use DecimalFormat to add the left padding.

PR #130, opened 2020-03-16
MAT-668. Made security banner work better in smaller windows.

cday-semanticbits:
Added horizontal scrolling and made the fonts smaller.


temp.json
```diff
@@ -0,0 +1,1395 @@
...
+  "measureId": "402803826d2fcd3c016d6ac98fcd0c07",
+  "modified": "2020-03-13T19:01:35.952Z",
+  "errorReason": "gov.cms.mat.fhir.services.exceptions.MatXmlMarshalException: org.exolab.castor.xml.MarshalException: ValidationException: Element 'title' occurs more than once. (parent class: mat.client.measure.ManageCompositeMeasureDetailModel\n location: /measureDetails/title{File: [not available]; line: 1; column: 283}",
+  "outcome": "INVALID_MEASURE_XML",
+  "conversionType": "VALIDATION",
+  "valueSetConversionResults": [],
+  "measureConversionResults": null,
+  "libraryConversionResults": [
+    {
+      "matId": "2ca880a170cb796a0170ceca8587008d",
```
serhii-ilin: you probably don't need it

src/main/java/mat/client/login/SecurityBannerModal.java
```diff
@@ -15,7 +15,14 @@
...
     private Button acceptButton = new Button("I Accept");
     private Button declineButton = new Button("Decline");
-    private String bodyHtml = "<br/><p>This warning banner provides privacy and security notices consistent with " +
+    private String scrollStyle = "width: 600px;" +
+                                 "height: 250px;" +
+                                 "font-size: x-small;" +
+                                 "padding-right: 10px;" +
+                                 "overflow-x: auto;" +
+                                 "overflow-x: auto;" +
+                                 "text-align: justify;";
```
adongare: this whole thing can go in MAT.css
```
.security-scrollbar {
  width: 600px;
  height: 250px;
  font-size: x-small;
  padding-right: 10px;
  overflow-x: auto;
  overflow-x: auto;
  text-align: justify;
}
```
then
`<div style="security-scrollbar">....</div>`

temp.json
cday-semanticbits: already removed

src/main/java/mat/client/login/SecurityBannerModal.java
cday-semanticbits: It certainly can, but what we have been doing is just coding it in with styles since we are not restyling. This is really customized just for this class. Other dialogs won't use this layout.

PR #132, opened 2020-03-16
MAT-668 Changed security banner to work better in lower res.

cday-semanticbits:
Also fixed compile issue with dictionary words.


pom.xml
```diff
@@ -15,18 +15,10 @@
         <java.version>1.8</java.version>
         <gwtVersion>2.8.2</gwtVersion>
         <googleGin>2.1.2</googleGin>
-        <org.springframework>5.0.6.RELEASE</org.springframework>
-        <org.springframework.security>5.0.6.RELEASE</org.springframework.security>
+        <org.springframework>5.0.15.RELEASE</org.springframework>
```
serhii-ilin: Hm... was it merged correctly?
I don't see those lines in POM.xml at all.
I think we removed those lines and we inherit from spring-boot-starter-parent

PR #142, opened 2020-03-19
MAT-750 Convert to FHIR - Success or Error Message for User

ghost:
https://jira.cms.gov/browse/MAT-750

image

### DIFF
--- a/src/main/java/mat/client/measure/ManageMeasurePresenter.java
+++ b/src/main/java/mat/client/measure/ManageMeasurePresenter.java
@@ -622,8 +622,7 @@ public class ManageMeasurePresenter implements MatPresenter, TabObserver {
 
     private void convertMeasureFhir(Result object) {
         if (showAlertAndReturnIfNotUMLSLoggedIn()) return;
-
-        GWT.log("Please wait. Conversion is in progress...");
+        logger.log(Level.INFO, "Please wait. Conversion is in progress...");
 
         if (!MatContext.get().getLoadingQueue().isEmpty()) {
             return;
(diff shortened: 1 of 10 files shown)


adongare (approved):

src/main/java/mat/server/service/FhirValidationReportService.java
```diff
@@ -28,19 +19,29 @@
...
 public class FhirValidationReportService {
 
+    private static final Log logger = LogFactory.getLog(FhirValidationReportService.class);
     private static final String DATE_FORMAT = "dd-MMM-YYYY";
     private static final String TIME_FORMAT = "hh:mm aa";
     private static final String NO_MEASURE_FOUND_ERROR = "The measure with that measure id does not exist.";
-    private static final String CONVERSION_SERVICE_ERROR = "An error occurred while validating the FHIR conversion. Please try again later.";
-    private static final Log logger = LogFactory.getLog(FhirValidationReportService.class);
+    private static final String CONVERSION_SERVICE_ERROR = "An error occurred while validating the FHIR conversion. Please try again later. If this continues please contact the MAT help desk.";
+    private static final String REPORT_FTL = "fhirvalidationreport/fhir_validation_report.ftl";
```
jkotanchik-SB: For consistency, could you move this to the properties file, please?

serhii-ilin: Doesn't make sense to me to move that to properties file. Those values are not configurable.

serhii-ilin: or, what do you mean?

jkotanchik-SB: Trying to avoid having hard coded template paths in multiple files. The email templates are currently set in the MAT.properties, but they're the only ones.

src/main/java/mat/client/measure/ManageMeasurePresenter.java
```diff
@@ -644,27 +643,24 @@ public void onFailure(Throwable caught) {
...
-                showFhirValidationReport(response.getFhirMeasure().getId());
+                showFhirValidationReport(response.getFhirMeasure().getId(), true);
                 displaySearch();
             }
         });
 
     }
 
-    private void showFhirValidationReport(String measureId) {
+    private void showFhirValidationReport(String measureId, boolean converted) {
```
jkotanchik-SB: If one is more common, converted or non-converted, would it make sense to add an overloaded method that defaults to the common argument.

src/main/java/mat/server/service/FhirValidationReportService.java
serhii-ilin: > Trying to avoid having hard coded template paths in multiple files.
Why? It doesn't make sense to me. We are not going to override that. No reason to make things more complex than needed.

src/main/java/mat/client/measure/ManageMeasurePresenter.java
jkotanchik-SB: If it is only used twice, then ignore me.

serhii-ilin: We could do that if there were more callers. We have just one method used in 2 places.

jkotanchik-SB (approved):

src/main/java/mat/client/measure/ManageMeasurePresenter.java
jkotanchik-SB: Not this particular method, but I saw you had to update some tests to pass "false" in a few places. Having to maintain tests and runtime code may make having a default overloaded method worthwhile.

serhii-ilin: It was a new test method to verify that the conversion throws an Exception when cqlSaved=false

PR #143, opened 2020-03-20
MAT-776 disable composite measures

ghost:
MAT-776
DISABLE Convert and FHIR Validation Buttons for Composite Measures

### DIFF
--- a/src/main/java/mat/server/service/impl/MatContextServiceUtil.java
+++ b/src/main/java/mat/server/service/impl/MatContextServiceUtil.java
@@ -196,7 +196,7 @@ public class MatContextServiceUtil implements InitializingBean {
                 !BooleanUtils.isTrue(measure.getIsCompositeMeasure());
     }
 
-    private static boolean isFhirMeasureValidatabe(Measure measure) {
+    private boolean isFhirMeasureValidatabe(Measure measure) {
         return measure.isDraft() && measure.isFhirMeasure();
     }
 
--- a/src/main/java/mat/server/service/impl/MatContextServiceUtil.java
+++ b/src/main/java/mat/server/service/impl/MatContextServiceUtil.java
@@ -186,18 +186,14 @@ public class MatContextServiceUtil implements InitializingBean {
     }
 
     public boolean isValidatable(Measure measure) {
-        return isFhirMeasureValidatabe(measure) || isQdmMeasureValidatable(measure);
+        return measure.isFhirMeasure() && measure.isDraft() || isQdmMeasureValidatable(measure);
     }
 
     private boolean isQdmMeasureValidatable(Measure measure) {
         return !measure.isDraft() &&
                 measure.isQdmMeasure() &&
                 isVersionEligibleForFhirValidation(measure.getQdmVersion(), measure.getReleaseVersion()) &&
-                !BooleanUtils.isTrue(measure.getIsCompositeMeasure());
-    }
-
-    private boolean isFhirMeasureValidatabe(Measure measure) {
-        return measure.isDraft() && measure.isFhirMeasure();
+                BooleanUtils.isNotTrue(measure.getIsCompositeMeasure());
     }
 
     public boolean isMeasureConvertible(Measure measure) {
@@ -206,7 +202,7 @@ public class MatContextServiceUtil implements InitializingBean {
         // 1.3 Pre-QDM measures cannot be converted. The button is disabled if the QDM version is before 5.5 or the MAT version is before 5.8.
         // Should be available for the owner or a super user
         String ownerId = measure.getOwner() == null ? null : measure.getOwner().getId();
-        return isConvertible(measure.getMeasureModel(), measure.isDraft(), measure.getQdmVersion(), measure.getReleaseVersion(), ownerId) && !BooleanUtils.isTrue(measure.getIsCompositeMeasure());
+        return isConvertible(measure.getMeasureModel(), measure.isDraft(), measure.getQdmVersion(), measure.getReleaseVersion(), ownerId) && BooleanUtils.isNotTrue(measure.getIsCompositeMeasure());
     }
 
     public boolean isConvertible(String modelType, boolean isDraft, String qdmVersion, String releaseVersion, String ownerId) {
--- a/src/main/java/mat/server/service/impl/MatContextServiceUtil.java
+++ b/src/main/java/mat/server/service/impl/MatContextServiceUtil.java
@@ -3,6 +3,7 @@ package mat.server.service.impl;
 import java.math.BigDecimal;
 import java.util.Objects;
 
+import org.apache.commons.lang3.BooleanUtils;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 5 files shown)


src/main/java/mat/server/service/impl/MatContextServiceUtil.java
```diff
@@ -184,17 +185,32 @@ public boolean isCurrentCQLLibraryDraftable(CQLLibraryDAO cqlLibraryDAO,
...
+    }
+
+    private boolean isQdmMeasureValidatable(Measure measure) {
+        return !measure.isDraft() &&
+                measure.isQdmMeasure() &&
+                isVersionEligibleForFhirValidation(measure.getQdmVersion(), measure.getReleaseVersion()) &&
+                !BooleanUtils.isTrue(measure.getIsCompositeMeasure());
+    }
+
+    private static boolean isFhirMeasureValidatabe(Measure measure) {
```
jkotanchik-SB: I don't follow why this method is static.

serhii-ilin: yeah.. I moved code from another helper class which was static. will remove

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
```diff
@@ -184,17 +185,32 @@ public boolean isCurrentCQLLibraryDraftable(CQLLibraryDAO cqlLibraryDAO,
         return isCurrentCQLLibraryEditable(cqlLibraryDAO, libraryId, false);
     }
 
+    public boolean isValidatable(Measure measure) {
+        return isFhirMeasureValidatabe(measure) || isQdmMeasureValidatable(measure);
```
jkotanchik-SB: Minor thing. Both sides will execute whenever it is a QDM measure. Maybe replace with either an if check on measure.isFhirMeasure() or a ternary operator.

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
```diff
@@ -184,17 +185,32 @@ public boolean isCurrentCQLLibraryDraftable(CQLLibraryDAO cqlLibraryDAO,
...
 
+    public boolean isValidatable(Measure measure) {
+        return isFhirMeasureValidatabe(measure) || isQdmMeasureValidatable(measure);
+    }
+
+    private boolean isQdmMeasureValidatable(Measure measure) {
+        return !measure.isDraft() &&
+                measure.isQdmMeasure() &&
+                isVersionEligibleForFhirValidation(measure.getQdmVersion(), measure.getReleaseVersion()) &&
+                !BooleanUtils.isTrue(measure.getIsCompositeMeasure());
```
jkotanchik-SB: Depending on how you feel about which is easier to read, BooleanUtils also has a isFalse() method (though I have not verified it is in the version of BooleanUtils we are using).

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
```diff
@@ -205,7 +221,7 @@ private boolean canUserConvert(String ownerId) {
         return isOwner || isSuperUser;
     }
 
-    private boolean isVersionEligibleForConversion(String qdmVersion, String matVersion) {
+    private boolean isVersionEligibleForFhirValidation(String qdmVersion, String matVersion) {
         return QDM_VER_BEFORE_CONVERSION.compareTo(asDecimalVersion(qdmVersion)) <= 0
                 && MAT_VER_BEFORE_CONVERSION.compareTo(asDecimalVersion(matVersion)) <= 0;
```
jkotanchik-SB: Instead of the class constants, could you the properties in MAT.properties for this comparison?
`mat.measure.current.release.version=v6.0`
and
`mat.measure.current.qdm.version=5.5`

jkotanchik-SB (approved):

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
serhii-ilin: yeah. it will. should not affect performance. it's not expensive.
not sure what solution you proposed. could you clarify?

like   return (measure.isFhir() && isFhirValidatable(measure)) || isQdmMeasureValidatable(measure)

but that will make code look less readable.. with minimum performance gains

cday-semanticbits: Something like:
measure.getMeasureModelModel() == FHIR ? isFhirMeasureVAlidatable(measure) : isQdmMeasureValidatable(measure);

jkotanchik-SB:
```
if (measure.isFhirMeasure()) {
  return isFhirMeasureValidatabe(measure);
}
return isQdmMeasureValidatable(measure);
```
Or
`return measure.isFhirMeasure() ? isFhirMeasureValidatabe(measure) : isQdmMeasureValidatable(measure);`

cday-semanticbits (approved):

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
serhii-ilin: ok

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
serhii-ilin: BooleanUtils.isFalse is checking specifically for False value. which means null will be treated as True

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
serhii-ilin: No. This is different. Those are threshold versions, not the current versions.

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
jkotanchik-SB: isFalse:
`Checks if a Boolean value is false, handling null by returning false.`.
From [javadocs](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/BooleanUtils.html#isFalse-java.lang.Boolean-)

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
jkotanchik-SB: Ah, ok. Thanks for clarifying!

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
serhii-ilin: well, I would really question any performance gains from that.
And I think that boolean expression is preferred to if/ternary one. 

@cday-semanticbits & @jkotanchik-SB 
How about measure.isFhirMeasure() && measure.isDraft() || isQdmMeasureValidatable(measure)

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
serhii-ilin: public static boolean isFalse(Boolean bool) {
        return Boolean.FALSE.equals(bool);
    }

serhii-ilin: Ah, got it. You suggested isNotTrue

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
cday-semanticbits: return measure.getMeasureModel() == FHIR ? isFhirMeasureVAlidatable(measure) : isQdmMeasureValidatable(measure); seems right to me but haven't looked at it closely

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
jkotanchik-SB: Did you want null to return true? I was assuming if measure.getIsCompositeMeasure() was null, then it wasn't a composite measure and should be considered false.

PR #145, opened 2020-03-21
MAT-771 Added new fhir global libs to liquibase. Changed conversion to make orchestration call with conversion instead of validation.

cday-semanticbits:
Added new fhir global libs to liquibase. Changed conversion to make orchestration call with conversion instead of validation.


src/main/java/mat/server/service/impl/FhirMeasureServiceImpl.java
```diff
@@ -61,23 +61,24 @@ public FhirConvertResultResponse convert(ManageMeasureSearchModel.Result sourceM
         ManageMeasureDetailModel sourceMeasureDetails = loadMeasureAsDetailsForCloning(sourceMeasure);
         dropFhirMeasureIfExists(sourceMeasureDetails);
 
-        ConversionResultDto conversionResult = validateSourceMeasureForFhirConversion(sourceMeasure, vsacGrantingTicket);
+        ConversionResultDto conversionResult = fhirOrchestrationGatewayService.convert(sourceMeasure.getId(), vsacGrantingTicket, sourceMeasure.isDraft());
```
serhii-ilin: validateSourceMeasureForFhirConversion is now probably unused and can be removed

There was some kind of a problem on ms side why we agreed to use validation instead of conversion. Maybe Duane recalls that. Hopefully, it's not an issue any more.

PR #146, opened 2020-03-23
[Bugfix] Fix from develop for user password check lazy init error

jkotanchik-SB:
[MAT-522] Fix hibernate lazy loading error (#91):
* Fixed hibernate lazy loading error when password history was being accessed from user entity outside of hibernate session. Replaced with call to DAO save.

* Fixed comments to describe updated day limits. Replaced on-demand import with individuals.

### DIFF
--- a/src/main/java/mat/server/service/impl/UserServiceImpl.java
+++ b/src/main/java/mat/server/service/impl/UserServiceImpl.java
@@ -72,17 +72,17 @@ import mat.shared.PasswordVerifier;
  */
 @Service
 public class UserServiceImpl implements UserService {
-	
+
 	/** The Constant logger. */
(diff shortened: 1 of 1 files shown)


jkotanchik-SB:
> Tough to see what was changed easily.
> Looks like it was just this: userPasswordHistoryDAO.save(passwordHistory);

Agreed. Not sure where all of the white space changes came from. There is only 2 changes, both in `addByUpdateUserPasswordHistory` at the bottom of the PR:

1. Replaced the password history save to use the userPasswordHistoryDAO instead of the User obj.
2. Moved the initial retrieval of the password history inside the if check, to avoid pulling from the DB when we're not saving an updated history.

PR #155, opened 2020-03-26
MAT-737: Removing Symantec VIP dependencies and code from develop

AndrewBird81:
Just stripping out all the VIP related code on develop


src/main/resources/MAT.properties
```diff
@@ -33,8 +33,7 @@ mat.qdm.default.expansion.id=Most Recent Code System Versions in VSAC
 #mat.vsac.server.url = https://vsac.nlm.nih.gov/vsac
 
 #MAT 2 factor authentication properties.
-#2FA_AUTH_CLASS=mat.server.twofactorauth.DefaultOTPValidatorForUser
-#2FA_AUTH_CLASS=mat.server.twofactorauth.vip.OTPValidatorForUser
+2FA_AUTH_CLASS=mat.server.twofactorauth.DefaultOTPValidatorForUser
```
jkotanchik-SB: I don't follow why this is being uncommented. Since we're removing 2FA from the MAT baseline, can this property be removed?

AndrewBird81: We still need the "default", do-nothing implementation for now. Once we go full-HARP/Okta, we can strip out more stuff. I wasn't trying to rock the boat too much with this PR, I just wanted to remove the VIP dependencies

PR #164, opened 2020-04-01
Consistent Case in Validation Report

RohitKandimalla:
Converting Severity to lower case and Capitalizing the first letter.


serhii-ilin:
Hm.. I was under impression that Michael wanted to implement that on MS side

PR #165, opened 2020-04-03
MAT-791 enhance rules for what can be converted

ghost:
https://jira.cms.gov/browse/MAT-791
1. Cannot convert if there is a QDM draft or an FHIR draft unrelated to the source measure in the family
2. Ask for confirmation if there is only one draft FHIR measure in the set and it was created from the QDM source.
3. Fix Hibernate mapping of the source measure.


cday-semanticbits (approved):
Looks good to me.

src/main/java/mat/server/MeasureLibraryServiceImpl.java
```diff
@@ -5931,8 +5936,7 @@ public String getHumanReadableForMeasureDetails(String measureId) {
             standardizeStartAndEndDate(measureInformationModel);
             humanReadableHTML = humanReadableGenerator.generate(measureInformationModel);
         } catch (Exception e) {
-            e.printStackTrace();
-            logger.info("Exception in getHumanReadableForMeasureDetails: " + e);
+            logger.info("Exception in getHumanReadableForMeasureDetails: " + e, e);
```
adongare, line 5939: Shouldn't this be `logger.error("msg", e)?`

adongare (approved):

serhii-ilin:
Needs for Chris's approval to merge.

src/main/java/mat/server/MeasureLibraryServiceImpl.java
serhii-ilin: oh, good catch

PR #167, opened 2020-04-08
MAT-390 performance issues

ghost:
Important changes:
1. Make sure that the whole search call is in the same transaction/session. Previously session load didn't work correctly. Measures, measure sets, etc were not cached in session because of different transaction scopes.
2. Increase batch size
3. Join what cannot be LAZY
4. skip 2nd sorting
5. equalsIngoreCase -> equals
6. Fetch all share levels with a single query instead of fetching them one-by-one in a cycle.
etc.
7. Some code style improvements and extracted common methods.
8. Fetch join measureDetails with measures
9. Feature flags were cached as a List. Every call to feature flags constructed a map. It didn't work well with a for loop over a list of measures.

This has drastically improved performance on my local from 25-30 sec to under 1 sec. That would be great to see how it improves performance in MatDev
There is still room for improvement.

### DIFF
--- a/src/main/java/mat/client/login/SecurityBannerModal.java
+++ b/src/main/java/mat/client/login/SecurityBannerModal.java
@@ -1,27 +1,28 @@
 package mat.client.login;
 
-import com.google.gwt.dom.client.Style;
-import com.google.gwt.user.client.ui.HTML;
 import org.gwtbootstrap3.client.ui.Anchor;
 import org.gwtbootstrap3.client.ui.Button;
 import org.gwtbootstrap3.client.ui.Modal;
-import org.gwtbootstrap3.client.ui.ModalHeader;
 import org.gwtbootstrap3.client.ui.ModalBody;
 import org.gwtbootstrap3.client.ui.ModalFooter;
+import org.gwtbootstrap3.client.ui.ModalHeader;
 import org.gwtbootstrap3.client.ui.gwt.FlowPanel;
 import org.gwtbootstrap3.client.ui.html.Div;
 import org.gwtbootstrap3.client.ui.html.Span;
+
+import com.google.gwt.dom.client.Style;
+import com.google.gwt.user.client.ui.HTML;
 
 public class SecurityBannerModal extends Modal {
     private Button acceptButton = new Button("I Accept");
     private Button declineButton = new Button("Decline");
     private String scrollStyle = "width: 930px;" +
-                                 "height: 260px;" +
-                                 "font-size: small;" +
-                                 "padding-right: 10px;" +
-                                 "overflow-x: auto;" +
-                                 "overflow-x: auto;" +
-                                 "text-align: justify;";
+            "height: 260px;" +
+            "font-size: small;" +
+            "padding-right: 10px;" +
+            "overflow-x: auto;" +
+            "overflow-x: auto;" +
+            "text-align: justify;";
     private String bodyHtml = "<div style=\"" + scrollStyle + "\"><br/><p>This warning banner provides privacy and security notices consistent with " +
             "applicable federal laws, directives, and other federal guidance for accessing this Government system, " +
             "which includes (1) this computer network, (2) all computers connected to this network, and (3) all " +
@@ -68,7 +69,7 @@ public class SecurityBannerModal extends Modal {
         ModalHeader modalHeader = new ModalHeader();
         Div headerDiv = new Div();
         headerDiv.add(new HTML("System Use Notification"));
-        headerDiv.getElement().getStyle().setProperty("font-size","small");
+        headerDiv.getElement().getStyle().setProperty("fontSize", "small");
         modalHeader.getElement().getStyle().setColor("white");
         modalHeader.getElement().getStyle().setBackgroundColor("blue");
         modalHeader.add(headerDiv);
@@ -87,7 +88,7 @@ public class SecurityBannerModal extends Modal {
         Div div = new Div();
         div.add(new Span("OMB No 0938-1236 | Expiration Date: 04/30/2017 (OMB Re-Certification Pending) | "));
         div.add(paperworkAnchor);
-        div.getElement().getStyle().setProperty("font-size","small");
+        div.getElement().getStyle().setProperty("fontSize", "small");
 
         modalBody.getElement().getStyle().setMarginRight(10, Style.Unit.PX);
         modalBody.getElement().getStyle().setMarginLeft(10, Style.Unit.PX);
@@ -96,8 +97,8 @@ public class SecurityBannerModal extends Modal {
         modalBody.add(new HTML(bodyHtml));
 
         ModalFooter modalFooter = new ModalFooter();
-        acceptButton.getElement().getStyle().setProperty("font-size","small");
-        declineButton.getElement().getStyle().setProperty("font-size","small");
+        acceptButton.getElement().getStyle().setProperty("fontSize", "small");
+        declineButton.getElement().getStyle().setProperty("fontSize", "small");
         modalFooter.add(acceptButton);
         modalFooter.add(declineButton);
 
@@ -117,7 +118,7 @@ public class SecurityBannerModal extends Modal {
         ModalHeader modalHeader = new ModalHeader();
         Div headerDiv = new Div();
         headerDiv.add(new HTML("Paperwork Reduction Act"));
-        headerDiv.getElement().getStyle().setProperty("font-size","small");
+        headerDiv.getElement().getStyle().setProperty("fontSize", "small");
         modalHeader.add(headerDiv);
         modalHeader.getElement().getStyle().setColor("white");
         modalHeader.getElement().getStyle().setBackgroundColor("blue");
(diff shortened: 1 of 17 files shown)


src/main/java/mat/model/User.java
```diff
@@ -209,7 +209,7 @@ public String getRootOID() {
         return "";
     }
 
-    @OneToOne(fetch = FetchType.LAZY, mappedBy = "user", cascade = CascadeType.ALL)
+    @OneToOne(fetch = FetchType.EAGER, mappedBy = "user", cascade = CascadeType.ALL)
```
serhii-ilin, line 212: One to one is basically EAGER if not instrumented, so joining it makes sense.

src/main/java/mat/client/login/SecurityBannerModal.java
```diff
@@ -1,27 +1,28 @@
...
     private Button declineButton = new Button("Decline");
     private String scrollStyle = "width: 930px;" +
-                                 "height: 260px;" +
-                                 "font-size: small;" +
-                                 "padding-right: 10px;" +
-                                 "overflow-x: auto;" +
-                                 "overflow-x: auto;" +
-                                 "text-align: justify;";
+            "height: 260px;" +
+            "font-size: small;" +
```
jkotanchik-SB, line 21: Should this be `fontSize`, or is there something different happening here than in the setProperty calls below?

serhii-ilin: As I understand if it's an html/style snippet, then a regular naming convention should be used.
 But setProperty accepts camel case names and it throws assertion errors in GWT/debug mode.

src/main/java/mat/dao/clause/impl/MeasureDAOImpl.java
```diff
@@ -555,19 +574,20 @@ private Predicate buildPredicateForMeasureSearch(String userId, CriteriaBuilder
 
     private List<MeasureShareDTO> getOrderedDTOListFromMeasureResults(MeasureSearchModel measureSearchModel, User user,
                                                                       List<Measure> measureResultList) {
-
         final ArrayList<MeasureShareDTO> orderedDTOList = new ArrayList<>();
 
         final Map<String, MeasureShareDTO> measureSetIdDraftableMap = new HashMap<>();
 
         final List<Measure> measureSets = getAllMeasuresInSet(measureResultList);
+        final Set<Measure> measureResultSet = measureResultList.stream().collect(Collectors.toSet());
```
jkotanchik-SB, line 582: Could `getAllMeasuresInSet` be refactored to return a `Set<Meassure>` instead of converting it?

src/main/java/mat/dao/clause/impl/MeasureDAOImpl.java
```diff
@@ -555,19 +574,20 @@ private Predicate buildPredicateForMeasureSearch(String userId, CriteriaBuilder
...
 
         for (final Measure measure : measureSets) {
             final MeasureShareDTO dto = extractDTOFromMeasure(measure);
             if (dto.isDraft()) {
                 measureSetIdDraftableMap.put(dto.getMeasureSetId(), dto);
             }
-            if (measureResultList.contains(measure)) {
+            if (measureResultSet.contains(measure)) {
+                measureResultSet.add(measure);
```
jkotanchik-SB, line 590: What's the intent in calling `add` when the measure is already contained in the set?

src/main/java/mat/dao/clause/impl/MeasureDAOImpl.java
serhii-ilin: It's probably confusing. But we don't need a set from measureSets. We need a set created from the original measureResultList, otherwise, there was for loop with arraylist.contains check. I'm not sure what requirements are covered in that logic so I don't feel confident to completely re-write it.

getAllMeasuresInSet returns a sorted list and order seems to be important.

Also, getAllMeasuresInSet is invoked 2 times on the same or similar data. I would prefer to drop the second call, but I cannot foresee the consequences :(

src/main/java/mat/dao/clause/impl/MeasureDAOImpl.java
```diff
@@ -854,6 +874,29 @@ public ShareLevel findShareLevelForUser(String measureId, String userID, String
...
+        NativeQuery query = session.createNativeQuery(SELECT_SHARE_LEVELS);
+        query.setParameter("userId", userID);
+        query.setParameter("measureSetIds", measureSetIds);
+        List<Object[]> list = query.getResultList();
+
+        Map<String, String> result = new HashMap<>();
+
+        list.stream().forEach(row -> {
+            result.put(String.valueOf(row[0]), String.valueOf(row[1]));
+        });
```
jkotanchik-SB, line 893: This could be done entirely in streams, though I don't know if it would be any more performant.
```
 Map<String, String> result = list.stream().collect(toMap(row -> String.valueOf(row[0]), row -> String.valueOf(row[1])));
```

src/main/java/mat/dao/clause/impl/MeasureDAOImpl.java
serhii-ilin: that's a good catch. it's not needed

src/main/java/mat/dao/clause/impl/MeasureDAOImpl.java
serhii-ilin: right.. just a code preference

src/main/java/mat/client/login/SecurityBannerModal.java
jkotanchik-SB: Understood, thanks for explaining!

src/main/java/mat/dao/clause/impl/MeasureDAOImpl.java
```diff
@@ -456,9 +471,13 @@ private Predicate getPredicateForShareList(String userId, String measureId, Hash
 
         if (isMeasureSetSearch(user, measureSearchModel)) {
             measureResultList = getAllMeasuresInSet(measureResultList);
+        } else {
+            measureResultList = sortMeasureList(measureResultList);
         }
```
jkotanchik-SB, line 476: nice catch preventing the double call to sortMeasureList!

src/main/java/mat/dao/clause/impl/MeasureDAOImpl.java
jkotanchik-SB: OK, thanks, I see where I got confused now.

jkotanchik-SB (approved):

src/main/java/mat/client/login/SecurityBannerModal.java
```diff
@@ -68,7 +69,7 @@ public SecurityBannerModal() {
         ModalHeader modalHeader = new ModalHeader();
         Div headerDiv = new Div();
         headerDiv.add(new HTML("System Use Notification"));
-        headerDiv.getElement().getStyle().setProperty("font-size","small");
+        headerDiv.getElement().getStyle().setProperty("fontSize", "small");
```
cday-semanticbits, line 72: Gotta love gwt. Some of the properties that they define you have to use camel case. For ones they don't its css name., font-size -> fontSize. Good catch.

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
```diff
@@ -68,24 +68,26 @@ public boolean isCurrentMeasureEditable(MeasureDAO measureDAO, String measureId)
...
         boolean isSuperUser = SecurityRole.SUPER_USER_ROLE.equals(userRole);
-        MeasureShareDTO dto = measureDAO.extractDTOFromMeasure(measure);
-        boolean isOwner = currentUserId.equals(dto.getOwnerUserId());
-        ShareLevel shareLevel = measureDAO.findShareLevelForUser(measureId,
-                currentUserId, dto.getMeasureSetId());
+        boolean isOwner = currentUserId.equals(measure.getOwner() == null ? null : measure.getOwner().getId());
         boolean isSharedToEdit = false;
-        if (shareLevel != null) {
-            isSharedToEdit = ShareLevel.MODIFY_ID.equals(shareLevel.getId());
+        if (shareLevelId != null) {
```
cday-semanticbits, line 84: isSharedToEdit = StringUtils.equals(shareLevelId,ShareLevel.MODIFY_ID) is cleaner imo. You don't need the if

cday-semanticbits (approved):
Looks good, had one small suggestion.

src/main/java/mat/server/service/impl/MatContextServiceUtil.java
serhii-ilin: Oh.. Good catch. That's a leftover of the old logic when it was an enum.

MODIFY_ID is never null, so a regular "equals" method is sufficient

cday-semanticbits: Ah right my bad, thought shareLevel was a string but it is an object.

PR #168, opened 2020-04-09
MAT-829 Allow Codeset/Valuset when editing library/measure cql.

cday-semanticbits:
This also includes the 2.0 version of the cql parser and many fixes to support fhir in MATs core code.
This also includes some fixes race conditions in CQLUtilityClass with the index static.


src/main/java/mat/cql/ConversionCqlToMatXml.java
```diff
@@ -0,0 +1,185 @@
...
+        vs.setId(newGuid());
+        vs.setName(name);
+        vs.setUuid(newGuid());
+        if (isOid(uri)) {
+            vs.setOid(parseOid(uri));
+        } else {
+            vs.setOid(uri);
+        }
+
+        // Nonsensical legacy stuff that has to be here for the gwt part to function at the moment.
```
jkotanchik-SB, line 100: 👍

PR #169, opened 2020-04-10
MAT-813 Disable Attribute drop down until a FHIR data type is selected

ghost:
https://jira.cms.gov/browse/MAT-813

In this story we're going to disable the FHIR attribute drop down in the Insert Attributes lightbox unless a data has been selected.


src/main/java/mat/client/cqlworkspace/InsertAttributeBuilderDialogBox.java
```diff
@@ -52,6 +54,9 @@
 import mat.model.clause.QDSAttributes;
 
 public class InsertAttributeBuilderDialogBox {
+
+    private static final Logger logger = Logger.getLogger("MAT");
```
adongare, line 58: Should this be `Logger.getLogger(InsertAttributeBuilderDialogBox.class)` to display this class name in logged message pattern?

adongare (approved):

src/main/java/mat/client/cqlworkspace/InsertAttributeBuilderDialogBox.java
serhii-ilin: It's a JUL logger (you cannot pass a class like that, only a name) and it's on a UI side.
What I found that GWT best practice is to use a root logger Logger.getLogger("") or a logger with a module name Logger.getLogger("MAT");

RohitKandimalla (approved):

PR #173, opened 2020-04-10
MAT-820 bug fix during create new Measure

spathiyil-sb:
When a new measure was being created, it was causing an error in the `mat.fhir.services` micro-service as it was expecting a `measure-details` row for each `measure` record

modify the logic for the `saveNewMeasure` to ensure that it creates a corresponding row in the `measure_details` table during the creation of a new `measure`.

### DIFF
--- a/src/main/java/mat/server/MeasureLibraryServiceImpl.java
+++ b/src/main/java/mat/server/MeasureLibraryServiceImpl.java
@@ -1764,6 +1764,13 @@ public class MeasureLibraryServiceImpl implements MeasureLibraryService {
                 measureSet.setId(UUID.randomUUID().toString());
                 measurePackageService.save(measureSet);
             }
+
+            if (pkg.getMeasureDetails() == null) {
+                MeasureDetails measureDetails = new MeasureDetails();
+                measureDetails.setMeasure(pkg);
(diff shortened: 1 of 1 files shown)


src/main/java/mat/server/MeasureLibraryServiceImpl.java
```diff
@@ -1764,6 +1764,13 @@ public final SaveMeasureResult saveOrUpdateMeasure(ManageMeasureDetailModel mode
                 measureSet.setId(UUID.randomUUID().toString());
                 measurePackageService.save(measureSet);
             }
+
+            if (pkg.getMeasureDetails() == null) {
+                MeasureDetails measureDetails = new MeasureDetails();
```
serhii-ilin, line 1769: Is there a way to cover that logic with unit test?

spathiyil-sb: I spent some time yesterday trying to write unit test for this, but given the current state of the code, it is exceedingly hard to write a good test.

serhii-ilin: ok, it was not a blocking comment.

PR #178, opened 2020-04-15
Feature/mat 728 convert cql library to fhir

RohitKandimalla:


src/main/java/mat/client/measure/service/FhirConvertResultResponse.java
```diff
@@ -2,13 +2,20 @@
 
 import com.google.gwt.user.client.rpc.IsSerializable;
 import mat.client.measure.ManageMeasureSearchModel;
+import mat.model.cql.CQLLibraryDataSetObject;
+import org.springframework.stereotype.Service;
 
+@Service
```
serhii-ilin, line 8: I don't think it's a service. It's a DTO object. The annotation is redundant.

src/main/java/mat/model/clause/CQLLibrary.java
```diff
@@ -332,6 +332,11 @@ public boolean isFhirMeasure() {
         return ModelTypeHelper.isFhir(getLibraryModelType());
     }
 
+    @Transient
```
serhii-ilin, line 335: Hm.. Is it a setter or a getter? It uses is* notation but accepts a parameter... Confusing.

src/main/java/mat/model/clause/CQLLibrary.java
```diff
@@ -332,6 +332,11 @@ public boolean isFhirMeasure() {
...
+    public boolean isFhirLibrary(String modelType) {
+        return ModelTypeHelper.isFhir(modelType);
+    }
+
     @Transient
     public boolean isQdmMeasure() {
         return ModelTypeHelper.isQdm(getLibraryModelType());
```
serhii-ilin, line 342: probably we should rename isQdmMeasure and isFhirMeasure methods to isQdmLibrary and isFhirLibrary

src/main/java/mat/model/clause/CQLLibrary.java
```diff
@@ -346,4 +351,22 @@ public void setFhirVersion(String fhirVersion) {
...
+        }
+        return xml;
+    }
+
+    public void setLibraryXMLAsByteArray(String xml) {
+        if (null != xml) {
+            byte[] xmlByteArr = xml.getBytes();
+            this.cqlXML = Hibernate.getLobCreator(HibernateConf.getHibernateSession()).createBlob(xmlByteArr);
+        } else {
+            this.cqlXML = null;
```
serhii-ilin, line 368: Did you try to get rid of an explicit LOB object? I believe that JPA should be able to handle byte array AS IS and a LOB instance will be created implicitly behind the scenes. That would reduce boiler plate code.

src/main/java/mat/server/CQLLibraryService.java
```diff
@@ -502,6 +509,43 @@ public void saveCQLLibraryExport(CQLLibrary cqlLibrary, String cqlXML) {
...
+        logger.info("checkLibraryForConversion  libraryId: " + sourceLibrary.getId() + " setId: " + sourceLibrary.getCqlSetId());
+        CheckForConversionResult result = new CheckForConversionResult();
+        List<CQLLibrary> draftLibraries = cqlLibraryDAO.getDraftLibraryBySet(sourceLibrary.getCqlSetId());
+        if (draftLibraries.isEmpty()) {
+            // If no drafts found we can proceed with conversion
+            result.setProceedImmediately(true);
+        } else {
+            // If the only draft is a FHIR draft created from the same source measure then ask for confirmation to override.
+            // UI cannot proceed with conversion if User didn't confirm or if there is/are other draft(s)
+            //TBD
```
serhii-ilin, line 523: TBD - probably you meant to add something here

src/main/java/mat/server/service/impl/FhirCqlLibraryServiceImpl.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    private void persistFhirLibrary(String loggedInUser, CQLLibrary existingLibrary, Optional<String> fhirCqlOpt) throws MatException, MarshalException, MappingException, ValidationException, IOException {
+        CQLLibrary newFhirLibrary = new CQLLibrary();
+        newFhirLibrary.setDraft(true);
+        newFhirLibrary.setName(existingLibrary.getName());
+        newFhirLibrary.setSetId(existingLibrary.getSetId());
+        newFhirLibrary.setOwnerId(existingLibrary.getOwnerId());
+        newFhirLibrary.setReleaseVersion(MATPropertiesService.get().getCurrentReleaseVersion());
+        newFhirLibrary.setQdmVersion(MATPropertiesService.get().getQdmVersion());
```
serhii-ilin, line 118: Hm.. I think I made QDM version property optional in MySQL. It's not required. But, instead of QDM version, you should use FHIR version property on an FHIR library.

src/main/java/mat/server/service/impl/FhirCqlLibraryServiceImpl.java
```diff
@@ -0,0 +1,158 @@
...
+        newFhirLibrary.setLibraryModelType(ModelTypeHelper.FHIR);
+
+        newFhirLibrary.setLibraryXMLAsByteArray(gneerateCqlXml(existingLibrary,fhirCqlOpt.get()));
+
```
serhii-ilin, line 122: typo 'gneerateCqlXml'

src/main/java/mat/server/service/impl/FhirCqlLibraryServiceImpl.java
RohitKandimalla: Sure I Will use FhirVersion instead of QdmVersion.

src/main/java/mat/server/service/impl/FhirCqlLibraryServiceImpl.java
```diff
@@ -0,0 +1,158 @@
...
+        FhirValidationStatus validationStatus = createValidationStatus(conversionResult);
+        fhirConvertResultResponse.setValidationStatus(validationStatus);
+
+        if (!fhirCqlOpt.isPresent()) {
+            // If there is no FHIR CQL, then we don't persist the Library. FHIR Library cannot be created.
+            throw new MatException("Your library cannot be converted to FHIR. Outcome: " + validationStatus.getOutcome() + " Error Reason: " + validationStatus.getErrorReason());
+        } else {
+            /*Create a new cql Library object and store it in DB*/
+            persistFhirLibrary(loggedInUser, existingLibrary, fhirCqlOpt);
```
serhii-ilin, line 88: at this point, it's known that fhirCqlOpt is present. you can pass unwrapped value with fhirCqlOpt.get

src/main/java/mat/server/service/impl/FhirCqlLibraryServiceImpl.java
```diff
@@ -0,0 +1,158 @@
...
+        return Optional.ofNullable(conversionResult.getLibraryConversionResults()).stream()
+                .flatMap(libConvRes -> libConvRes.stream())
+                .map(cqlLibRes -> cqlLibRes.getCqlConversionResult())
+                .filter(el -> el != null)
+                .map(el -> el.getFhirCql())
+                .filter(StringUtils::isNotBlank)
+                .findFirst();
+    }
+
+    private void persistFhirLibrary(String loggedInUser, CQLLibrary existingLibrary, Optional<String> fhirCqlOpt) throws MatException, MarshalException, MappingException, ValidationException, IOException {
```
serhii-ilin, line 111: I recall there were some issues with 'persistFhirMeasure'. Its steps were not executed within the same transaction and the change was not atomic. An error in the middle could cause artifacts  - like a persisted FHIR measure with QDM CQL ..
So, you might want to double-check if all the steps are executed within the same transaction in  persistFhirLibrary.

serhii-ilin: Maybe not needed in that case.

src/main/java/mat/dao/clause/CQLLibraryDAO.java
```diff
@@ -20,6 +20,8 @@
 
 	String findMaxVersion(String setId, String ownerId);
 
+    CQLLibrary findByNameAndVersion(String name, double version, int revisionNumber);
```
serhii-ilin, line 23: formatting :)
the line is not aligned the same way as other methods

src/main/java/mat/dao/clause/CQLLibraryDAO.java
```diff
@@ -46,5 +48,7 @@
 
 	String getLibraryNameIfDraftAlreadyExists(String setId);
 
-	CQLLibrary findByNameAndVersion(String name,double releaseVersion, int revisionNumber);
+    List<CQLLibrary> getDraftLibraryBySet(String cqlSetId);
```
serhii-ilin, line 51: something with alignment. do you have any custom intellij settings?

src/main/java/mat/model/clause/CQLLibrary.java
RohitKandimalla: Those 2 methods are not created now. and isFhirMeasure method is being used in CqlLibraryService class

src/main/java/mat/server/CQLLibraryService.java
RohitKandimalla: I had one but forgot to remove the comment.

src/main/java/mat/dao/clause/CQLLibraryDAO.java
RohitKandimalla: Not really and this looks fine in IDE

PR #181, opened 2020-04-16
Created ConversionResultDto in MAT

RohitKandimalla:


serhii-ilin:
Let's cleanup pom dependencies then..

src/main/java/mat/DTO/fhirconversion/CqlConversionResult.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.List;
+import java.util.Set;
+
+@Data
+@NoArgsConstructor
+public class CqlConversionResult {
+    ConversionType type;
+    Boolean result;
+    List<String> errors = new ArrayList();
+    @Transient
```
cday-semanticbits, line 18: I would remove these @Transient annotations. 
I don't even think it is even the right annotation to use here on the ms side they should have used the jackson one. I know the fhirCql still comes across.even with it sent as @Transient.

src/main/java/mat/DTO/fhirconversion/ConversionOutcome.java
```diff
@@ -0,0 +1,26 @@
+package mat.DTO.fhirconversion;
+
+import lombok.NoArgsConstructor;
+
+@NoArgsConstructor
+public enum ConversionOutcome {
```
cday-semanticbits, line 6: I would make this a string instead of an enum. It makes it more flexible on the MS side this way. They can add whatever we want and we don't have to update this enum every time they do it because it breaks.

src/main/java/mat/DTO/fhirconversion/CqlConversionResult.java
RohitKandimalla: Sure, How about access modifiers for other few variables?

cday-semanticbits: ah Codacy ... I just make them all private when I see that

RohitKandimalla: ok, will make them private

PR #188, opened 2020-04-21
MAT-832 Fixing value-set tab in measures and libs.

cday-semanticbits:
Fixed value-set tab for FHIR measures.
Updated global libs to comment out Bryn's suggestion from an email.
Added in a lot of logging in measure workspace so it's easier to track down GWT errors.


src/main/java/mat/client/cqlworkspace/CQLMeasureWorkSpacePresenter.java
```diff
@@ -1912,7 +1932,14 @@ protected String getCurrentModelType(){
 		return MatContext.get().getCurrentMeasureModel();
 	}
 
+	protected boolean isFhirModelType() {
+		boolean result = !"QDM".equals(getCurrentModelType());
```
serhii-ilin, line 1936: Legacy pre-QDM measures can have model type  null/empty

cday-semanticbits: Good catch, will change that to "FHIR".equals instead of !"QDM".equals so it handles all those cases too.

cday-semanticbits: Changed it to use the MatContext.get().isFhirModelType method I added that uses "FHIR".equals and removed that method in wspresenter.

PR #198, opened 2020-04-30
MAT-804 display cql library validation report

spathiyil-sb:
Log into MAT. 
Log into UMLS.
Click on CQL Library tab at the top.
Search for a QDM library.
Click `Convert to FHIR` button.

Unlike Measures, when conversion errors were happening, it was not displaying the errors in a separate tab. With this change, it will open a tab and display a validation report for the CQL conversion errors for a CQL library.

add an extra parameter named `type` to the servlet call that opens a tab to display the validation report. The valid value for `type` is `measure` and `library`

https://jira.cms.gov/browse/MAT-804


src/main/java/mat/server/fhirvalidation/FhirValidationReportServlet.java
```diff
@@ -14,50 +16,74 @@
...
+    }
+
+    /**
+     * Given sessionId, delegates to VsacApiService to fetch the VSAC ticket
+     * @param sessionId
+     * @return vsac Ticket
+     * @throws MatException
+     */
+    private String getVsacTicket(String sessionId) throws MatException {
+        var vsacTicketInformation = getVascApiService().getTicketGrantingTicket(sessionId);
```
RohitKandimalla, line 64: To Validate/convert Standalone Cql library, we don't need Vsac token. Can you Confirm with Michael ?

cday-semanticbits: Right now we don't, but in stories yet to come we will need it again. Please leave it in there.

RohitKandimalla: Gotcha

PR #199, opened 2020-04-30
[MAT-898] Updated logout flow to prevent Okta session retension

jkotanchik-SB:
## [MAT-898: HARP OKTA: Logging out does not end the OKTA session](https://jira.cms.gov/browse/MAT-898)

**Issue:** The MAT Logout operations and the Okta logout API were attempting to re-direct to the login page, creating a race condition on whether or not the Okta logout would complete before being cancelled by the redirect.

**Solution:** Update the logout flow to wait to redirect until all logout operations are complete and move the Okta logout call to the end to ensure it's not being impacted.

1. Revoke Access Token.
2. Complete MAT logout operations.
3. Destroy browser Okta session by calling Otka logout endpoint.
4. Wait 1 second, then re-direct to login page.

Previously, there step 2 & 3 were swapped and there was no wait before the re-direct.

### DIFF
--- a/src/main/java/mat/client/Mat.java
+++ b/src/main/java/mat/client/Mat.java
@@ -172,7 +172,7 @@ public class Mat extends MainLayout implements EntryPoint, Enableable, TabObserv
         public void onFailure(Throwable throwable) {
             logger.log(Level.SEVERE, "LoginService::initSession -> onFailure: " + throwable.getMessage(), throwable);
             if (throwable.getMessage().contains("MAT_ACCOUNT_REVOKED_LOCKED")) {
-                String msg = "Harp UserName doesn't match existing MAT HARP_ID, redirecting to access support page";
+                String msg = "HARP User's access to the MAT has been revoked, redirecting to access support page";
                 logger.log(Level.SEVERE, msg);
                 MatContext.get().recordTransactionEvent(null, null, "MAT_ACCOUNT_REVOKED_LOCKED", msg, 0);
             } else if (throwable.getMessage().contains("HARP_ID_NOT_FOUND")) {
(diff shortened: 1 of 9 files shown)


src/main/java/mat/client/Mat.java
```diff
@@ -649,26 +647,19 @@ private void harpOktaLogout(String redirectTo) {
...
             }
 
             @Override
             public void onSuccess(String harpUrl) {
                 logger.log(Level.INFO, "HarpService::getHarpUrl -> onSuccess");
-                harpLogout(harpUrl);
                 MatContext.get().getSynchronizationDelegate().setLogOffFlag();
-                MatContext.get().handleSignOut("SIGN_OUT_EVENT", redirectTo);
-                removeOktaTokens();
+                MatContext.get().handleSignOut("SIGN_OUT_EVENT", null);
```
serhii-ilin, line 656: if it's set to null, then redirectUrl will not work to Harp Support page

jkotanchik-SB: Understood, I'll fix it shortly

src/main/java/mat/client/Mat.java
```diff
@@ -649,26 +647,19 @@ private void harpOktaLogout(String redirectTo) {
...
+                harpLogout(harpUrl);
+                redirectToLogin();
             }
         });
     }
 
-    private void removeOktaTokens() {
-        logger.log(Level.INFO, "removeOktaTokens");
```
serhii-ilin, line 667: Interesting..

src/main/java/mat/client/Mat.java
jkotanchik-SB: Fixed.

src/main/java/mat/client/Mat.java
jkotanchik-SB: Removing the tokens didn't seem to have any affect on logout functionality. I've added the method back to avoid leaving the tokens in memory as a precaution.

PR #208, opened 2020-05-02
 MAT-887: Lots of styling of the login footer

AndrewBird81:


war/HarpLogin.html
```diff
@@ -51,15 +143,98 @@
...
+</form>
+
+<footer class="footer">
+    <div class="container">
+        <div class="row">
+            <div class="col-md-10 footer-content">
+                <div>
+                    <a title="CMS home page" tabindex="-1" href="http://www.cms.gov/" target="_blank">
+                        <h4 class="screen-reader">CMS.gov</h4>
+                        <img class="mat-logo" alt="CMS.gov" tabIndex="0" src="./images/MAT-logo_white.svg">
```
jkotanchik-SB, line 160: Alt is inaccurate.

AndrewBird81: The logo is that CMS.gov | MEASURE AUTHORING TOOL. Either way, I'll update it later. This is fine for now

PR #216, opened 2020-05-05
[MAT-995] Swap login urls & increment version

jkotanchik-SB:
- Renamed `Login.html` -> `LegacyLogin.html`
- Renamed `HarpLogin.html` -> `Login.html`
- Excluded `LegacyLogin.html` from war packaging.
- Incremented release version in `MAT.properties` and `pom.xml`
- Updated logout iframe styling to hide it better.


war/WEB-INF/applicationContext-security.xml
```diff
@@ -22,16 +22,16 @@
     	<csrf disabled="true"/>
         <custom-filter position="PRE_AUTH_FILTER" ref="preAuthnHarpFilter"/>
         <intercept-url pattern="/HarpSupport.html" access="permitAll()"/>
-        <intercept-url pattern="/HarpLogin.html" access="permitAll()"/>
+        <intercept-url pattern="/Login.html" access="permitAll()"/>
```
serhii-ilin, line 25: Should we see legacylogin listed somewhere here? Or we just leave it for development only

jkotanchik-SB: Development only. If it is made available in PROD, we'd be non-compliant.

serhii-ilin: Thanks for the clarification

PR #234, opened 2020-05-12
MAT-944 Partial fix for Success with errors showing when there are no errors.

cday-semanticbits:
Fixed this to work in the MAT side with this change. Still needs some additional MS work to work fully though. Checking it in so it doesn't get lost.


war/WEB-INF/applicationContext-mail.xml
```diff
@@ -10,7 +10,7 @@
 >
 	
     <bean id="mailSender" class="org.springframework.mail.javamail.JavaMailSenderImpl">
-        <property name="host" value="localhost"/>
+        <property name="host" value="172.23.20.122"/>
```
serhii-ilin, line 13: what is this IP?

PR #245, opened 2020-05-26
MAT-1012 qdm version/packaging bug

ghost:


src/main/java/mat/model/cql/CQLIncludeLibrary.java
```diff
@@ -123,11 +126,12 @@ public boolean equals(Object arg0) {
             return false;
         }
 
+        // (cqlIncludeLibrary.libraryModelType == libraryModelType || libraryModelType != null && libraryModelType.equals(cqlIncludeLibrary.libraryModelType))
         if (cqlIncludeLibrary.cqlLibraryId.equals(cqlLibraryId) &&
                 cqlIncludeLibrary.aliasName.equals(aliasName) &&
                 cqlIncludeLibrary.cqlLibraryName.equals(cqlLibraryName) &&
                 cqlIncludeLibrary.version.equals(version) &&
-                cqlIncludeLibrary.libraryModelType.equals(libraryModelType)) {
+                (Objects.equals(libraryModelType, cqlIncludeLibrary.libraryModelType))) {
```
serhii-ilin, line 134: We had a NullPointer here

src/main/java/mat/model/cql/CQLIncludeLibrary.java
```diff
@@ -25,7 +28,7 @@
 
     private String measureId;
 
-    private String libraryModelType;
+    private String libraryModelType = ModelTypeHelper.QDM;
```
serhii-ilin, line 31: The library's model type is not persisted in Cator'x model XML. 
We should probably change that.

src/main/java/mat/model/cql/CQLIncludeLibrary.java
serhii-ilin: libraryModelType was null

cday-semanticbits (approved):
Nice clean up as well as fixing the issue.

src/main/java/mat/server/service/impl/MeasurePackageServiceImpl.java
```diff
@@ -556,8 +556,10 @@ private void createAndSaveExportsAndArtifacts(MeasureExport export, boolean shou
         measure.setExportedDate(new Date());
         measureDAO.save(measure);
         measureExportDAO.save(export);
-        //New for FHIR we create a CQL_LIBRARY in draft state for the measure so microservices can find it.
-        measureLibraryService.exportDraftCQLLibraryForMeasure(measure);
+        if (measure.isFhirMeasure()) {
```
serhii-ilin, line 559: @cday-semanticbits 
JFYI - this custom FHIR code caused issues to QDM packaging so we now run it for FHIR measured only

src/main/java/mat/server/CQLUtilityClass.java
```diff
@@ -351,7 +351,7 @@ public static String getXMLFromCQLModel(CQLModel cqlModel) {
             marshaller.marshal(cqlModel);
             xml = stream.toString();
         } catch (MarshalException | ValidationException | IOException | MappingException e) {
-            e.printStackTrace();
+            logger.error("Error in  getXMLFromCQLModel: " + e.getMessage(), e);
```
adongare, line 354: Message will be there in stacktrace. I think no need of `e.getMessage`

adongare (approved):

src/main/java/mat/server/service/impl/MeasurePackageServiceImpl.java
cday-semanticbits: Good catch, should have been for FHIR only.

PR #266, opened 2020-06-02
Feature/hcqis mat sbx dev

BrDonohue:
This is to update dockerfile, jenkinsfile context and entrypoint for building mat on hcqis dev.


BrDonohue:
rethinking this

PR #270, opened 2020-06-08
[MAT-1058] Use Okta's userinfo endpoint to retrieve user claims (MASTER)

jkotanchik-SB:
Use claims from userinfo Instead of relying on the id token when updating user details in MAT. The response from userinfo is easier to parse and more consistent as it is not affected by okta custom authorization server rules.

- Added new `HarpConstant`s for user's given name, middle name, and family name. (Previously, we only had easy access to the user's fullname on the ID Token).

- Added operation to HarpService to call Okta's `/userinfo` endpoint and parse the results into a HashMap with `HarpConstant`s as keys.

- Removed `Mat.java`'s parsing user details from the ID Token. Added call to `HarpService.getUserInfo`, which happens after validating the access token, since it is required, and the result is then saved to `MatContext` so the existing session flow can continue.

- Updated `LoginCredentialService.updateUserDetails` to use the Map values instead of parsing the first and last names from a single string.

### DIFF
--- a/src/main/java/mat/server/HarpServiceImpl.java
+++ b/src/main/java/mat/server/HarpServiceImpl.java
@@ -86,7 +86,7 @@ public class HarpServiceImpl extends SpringRemoteServiceServlet implements HarpS
         acceptList.add(MediaType.APPLICATION_JSON);
         headers.setAccept(acceptList);
         headers.setContentType(MediaType.APPLICATION_FORM_URLENCODED);
-        HttpEntity<TokenIntrospect> request = new HttpEntity<>(headers);
+        HttpEntity<Void> request = new HttpEntity<>(headers);
         Map<String, Object> uriVariables = new HashMap<>();
         uriVariables.put("clientId", getHarpClientId());
         uriVariables.put("hint", "access_token");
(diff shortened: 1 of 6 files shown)


src/main/java/mat/server/HarpServiceImpl.java
```diff
@@ -114,43 +123,40 @@ public String getHarpClientId() {
...
-//    }
+    @Override
+    public Map<String, String> getUserInfo(String accessToken) {
+        HttpHeaders headers = new HttpHeaders();
+        List<MediaType> acceptList = new ArrayList<>();
+        acceptList.add(MediaType.APPLICATION_JSON);
+        headers.setAccept(acceptList);
+        headers.setContentType(MediaType.APPLICATION_FORM_URLENCODED);
+        headers.set("Authorization", "Bearer " + accessToken);
+        HttpEntity<TokenIntrospect> request = new HttpEntity<>(headers);
```
serhii-ilin, line 134: Hm... I'm confused. 
HttpEntity of TokenIntrospect type but accepts only headers and there is no body instance of TokenIntrospect?

jkotanchik-SB: Probably because it's wrong.

PR #287, opened 2020-06-16
MAT-1121	POC to expand CQL editor.

serhii-ilin:


serhii-ilin:
Known issues:
1.  AceEditor is difficult to resize. Additional research should be done to adjust the line width accordingly. Most likely it should be done via JSNI. There are several div containers that get fixed width/height assigned somehow.
2. There are native browser events (on full-screen enter, on full-screen exit) that need to be handled, otherwise ESC key is not working properly.
3. The side panel splitter can be multiplied (if you navigate MAT back and forth you can see several collapses buttons). There is a chance it's an existing issue. Also, if the sidebar is collapsed there is no easy way for CQL Editor to acquire that space. The layout is implemented as a several nested tables with assigned width/height. So this functionality needs for refactoring 
4. I didn't test the approach with standalone CQL editor.

PR #314, opened 2020-07-10
Implemented MAT-268 and MAT-408. Adding feature flag to measure/lib share button.

cday-semanticbits:


src/main/java/mat/client/shared/CQLibraryGridToolbar.java
```diff
@@ -135,11 +136,15 @@ public void updateOnSelectionChanged(CQLLibraryDataSetObject selectedItem) {
...
         }
 
+
         shareButton.setText(SHARE_TEXT);
-        shareButton.setEnabled(selectedItem.isSharable());
         shareButton.setIcon(IconType.SHARE_SQUARE);
         shareButton.setTitle(CLICK_TO_SHARE_TITLE);
-
+        if (ModelTypeHelper.isFhir(selectedItem.getLibraryModelType())) {
+            shareButton.setEnabled(selectedItem.isSharable() && MatContext.get().getFeatureFlagStatus("FhirShare"));
```
RohitKandimalla, line 144: Better to use "FeatureFlagConstant"

cday-semanticbits: changed to that

PR #541, opened 2021-02-05
Removed Cql Editor Tooltip

RohitKandimalla:
This change removes the Tooltip "CQL Library Editor" 
Technically removing a Title tag, Tested with Google Chrome screen Reader, not impacted

https://jira.cms.gov/browse/MAT-2464

### DIFF
--- a/src/main/java/mat/client/cqlworkspace/shared/CQLEditorPanel.java
+++ b/src/main/java/mat/client/cqlworkspace/shared/CQLEditorPanel.java
@@ -29,7 +29,6 @@ public class CQLEditorPanel extends Composite {
 		this.isReadOnly = isReadOnly;
 		this.editor = new CQLEditor(isReadOnly);
 		this.editor.getElement().getElementsByTagName("textarea").getItem(0).setTitle(text);
-		this.editor.getElement().setTitle(text);
 		initWidget(buildWidget());
 	}
 	


src/main/java/mat/client/cqlworkspace/shared/CQLEditorPanel.java
```diff
@@ -29,7 +29,6 @@ public CQLEditorPanel(String name, String text, boolean isReadOnly) {
 		this.isReadOnly = isReadOnly;
 		this.editor = new CQLEditor(isReadOnly);
 		this.editor.getElement().getElementsByTagName("textarea").getItem(0).setTitle(text);
```
jkotanchik-SB, line 31: Does this need an aria label or something for screenreaders?

jkotanchik-SB (approved):

src/main/java/mat/client/cqlworkspace/shared/CQLEditorPanel.java
RohitKandimalla: The parent tag already contains an aria-label
aria-label="You are entering a CQL Editor. To exit the editor backward press shift and up arrow. To exit the editor forward press shift and down arrow."

greenemcg (approved):

PR #664, opened 2021-04-29
Added a check to see if the revision number is '0' instead of '000'

RohitKandimalla:
MAT-2832
## Description

The revision number of Cql library is retrieved as '0' and is stored in measure XML, but the actual cql library will have '000' as the revision number. This mismatch is causing an issue while replacing the included libraries from helper UI. 
Solution: While retrieving the cql Library, if the revision number is '0' then we are returning '000' instead So that it could be saved as '000' in measure Xml.

## JIRA Ticket

[MAT-2832](https://jira.cms.gov/browse/MAT-2832)
## Checklist

- [ ] None applicable

### DIFF
--- a/src/main/java/mat/server/CQLLibraryService.java
+++ b/src/main/java/mat/server/CQLLibraryService.java
@@ -249,7 +249,7 @@ public class CQLLibraryService extends SpringRemoteServiceServlet implements CQL
         dataSetObject.setReleaseVersion(cqlLibrary.getReleaseVersion());
         dataSetObject.setQdmVersion(cqlLibrary.getQdmVersion());
         dataSetObject.setFhirVersion(cqlLibrary.getFhirVersion());
-        if (cqlLibrary.getRevisionNumber() == null) {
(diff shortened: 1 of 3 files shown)


src/main/java/mat/server/CQLLibraryService.java
```diff
@@ -249,7 +249,7 @@ public CQLLibraryDataSetObject extractCQLLibraryDataObject(CQLLibrary cqlLibrary
         dataSetObject.setReleaseVersion(cqlLibrary.getReleaseVersion());
         dataSetObject.setQdmVersion(cqlLibrary.getQdmVersion());
         dataSetObject.setFhirVersion(cqlLibrary.getFhirVersion());
-        if (cqlLibrary.getRevisionNumber() == null) {
+        if (cqlLibrary.getRevisionNumber() == null || cqlLibrary.getRevisionNumber().equalsIgnoreCase("0")) {
```
RohitKandimalla, line 252: This is the fix.

jkotanchik-SB (approved):

PR #667, opened 2021-04-30
Added check for Continuous Variable Patient Based to only have one Me…

greenemcg:
MAT-2828

## Description
Added check for Continuous Variable Patient Based to only have one Me…

## Checklist
Only change required, I tested all conditions per the story and could the system validated correctly per my tests.

### DIFF
--- a/src/main/java/mat/server/service/impl/PatientBasedValidator.java
+++ b/src/main/java/mat/server/service/impl/PatientBasedValidator.java
@@ -32,7 +32,7 @@ import java.util.Map;
 @Component
 public class PatientBasedValidator {
 	
-		private static final String SCORING_CONTINUOUS_VARIABLE = "Continuous Variable";
+	private static final String SCORING_CONTINUOUS_VARIABLE = "Continuous Variable";
 
 	private static final String SCORING_RATIO = "Ratio";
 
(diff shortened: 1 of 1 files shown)


src/main/java/mat/server/service/impl/PatientBasedValidator.java
```diff
@@ -145,7 +148,11 @@
 				createExpressionsToBeCheckedData(expressionPopMap, exprList, name, definitionName);
 			}
 		}
-		
+
+		if( scoringType.equalsIgnoreCase(SCORING_CONTINUOUS_VARIABLE) && measureObservationCount > 1) {
+			errorMessages.add(SCORING_CONTINUOUS_VARIABLE +  " measures can only have one measure observation.");
```
jkotanchik-SB, line 153:
```suggestion
			errorMessages.add(SCORING_CONTINUOUS_VARIABLE +  " measures may only have one Measure Observation per Measure Group");

```

jkotanchik-SB, line 153:
```suggestion
			errorMessages.add(SCORING_CONTINUOUS_VARIABLE +  " measures can only have one Measure Observation per Measure Grouping.");

```

jkotanchik-SB (approved):

PR #693, opened 2021-06-10
Additional Custom Error Pages

jkotanchik-SB:
## Description
Set link to prod home page in error pages instead of using relative link to /Mat.html as it can readily fail if the user has navigated to an unsupported path.

Map common 400/500 errors to custom error page. Do not use the defaut tomcat error page as it includes stack trace and server version information that could be used by attackers.

## Checklist

### Custom Error Page:
Screen Shot 2021-06-10 at 11 36 01 AM


sb-benohe (approved):
LGTM, This should fix it, I'll see if there are any more errors we missed.

PR #774, opened 2022-07-08
Fix for performance slowdown observed during packaging and versioning…

RohitKandimalla:
… of FHIR measures

## Checklist

- [ ] None applicable

### DIFF
--- a/src/main/java/cqltoelm/parsers/MATCQL2ELMListener.java
+++ b/src/main/java/cqltoelm/parsers/MATCQL2ELMListener.java
@@ -414,13 +414,15 @@ public class MATCQL2ELMListener extends cqlBaseListener {
         // updated again if need be.
         if (element instanceof IncludeDef) {
             IncludeDef def = (IncludeDef) element;
-            libraries.add(def.getPath() + "-" + def.getVersion() + "|" + def.getLocalIdentifier());
             graph.addEdge(currentContext, def.getPath() + "-" + def.getVersion() + "|" + def.getLocalIdentifier());
             libraryAccessor = def;
-            try {
-                parseChildLibraries(def);
-            } catch (IOException e) {
-                e.printStackTrace();
+            if (!libraries.contains(def.getPath() + "-" + def.getVersion() + "|" + def.getLocalIdentifier())) {
+                libraries.add(def.getPath() + "-" + def.getVersion() + "|" + def.getLocalIdentifier());
+                try {
+                    parseChildLibraries(def);
+                } catch (IOException e) {
+                    e.printStackTrace();
+                }
             }
         } else if (element instanceof CodeDef) {
             codes.add(formattedIdentifier);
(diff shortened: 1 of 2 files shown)


src/main/java/cqltoelm/parsers/MATCQL2ELMListener.java
```diff
@@ -414,13 +414,15 @@ private Element resolve(String identifier, TranslatedLibrary library) {
...
             IncludeDef def = (IncludeDef) element;
-            libraries.add(def.getPath() + "-" + def.getVersion() + "|" + def.getLocalIdentifier());
             graph.addEdge(currentContext, def.getPath() + "-" + def.getVersion() + "|" + def.getLocalIdentifier());
             libraryAccessor = def;
-            try {
-                parseChildLibraries(def);
-            } catch (IOException e) {
-                e.printStackTrace();
+            if (!libraries.contains(def.getPath() + "-" + def.getVersion() + "|" + def.getLocalIdentifier())) {
+                libraries.add(def.getPath() + "-" + def.getVersion() + "|" + def.getLocalIdentifier());
```
jkotanchik-SB, line 420: How about moving the `add` inside the try/catch block, right after `parseChildLibraries` to ensure we're only adding the library if the method is successfully. Though, we don't do anything different in the error case 🤣

src/main/java/cqltoelm/parsers/MATCQL2ELMListener.java
```diff
@@ -414,13 +414,15 @@ private Element resolve(String identifier, TranslatedLibrary library) {
...
+                try {
+                    parseChildLibraries(def);
```
jkotanchik-SB, line 422:
```suggestion
                try {
                    parseChildLibraries(def);
                    libraries.add(def.getPath() + "-" + def.getVersion() + "|" + def.getLocalIdentifier());
```
How about moving the `add` inside the try/catch block, right after `parseChildLibraries` to ensure we're only adding the library if the method is successfully. Though, we don't do anything different in the error case

src/main/java/mat/server/CQLServiceImpl.java
```diff
@@ -1642,13 +1642,7 @@ public GetUsedCQLArtifactsResult getUsedCQlArtifacts(String xml) {
...
         }
 
-        String cqlString  = CQLUtilityClass.getCqlString(cqlModel, "").getLeft();
-        SaveUpdateCQLResult cqlResult;
-        if (cqlModel.isFhir()) {
-            cqlResult = parseFhirCQLForErrors(cqlModel, cqlString);
-        } else {
-            cqlResult = CQLUtil.parseQDMCQLLibraryForErrors(cqlModel, getCqlLibraryDAO(), exprList);
-        }
+        SaveUpdateCQLResult cqlResult = CQLUtil.parseQDMCQLLibraryForErrors(cqlModel, getCqlLibraryDAO(), exprList);
```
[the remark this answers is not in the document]
RohitKandimalla: yup no longer needed

src/main/java/cqltoelm/parsers/MATCQL2ELMListener.java
RohitKandimalla: Yup it doesn't matter

jkotanchik-SB: Yes, pls.

jkotanchik-SB (approved):

PR #967, opened 2024-01-26
MAT-6580 Place Single Transfer Button Logic Behind Feature Flag - QDM ONLY

adongare:
…

## JIRA Ticket

[MAT-6580](https://jira.cms.gov/browse/MAT-6580) Place Single Transfer Button Logic Behind Feature Flag - QDM ONLY

## Checklist

- [x] Tests are included and test edge cases.
- [x] Tests have been run locally and pass.

- [ ] None applicable

### DIFF
--- a/src/main/java/mat/client/shared/MessageDelegate.java
+++ b/src/main/java/mat/client/shared/MessageDelegate.java
@@ -211,7 +211,7 @@ public class MessageDelegate {
     private static final String MEASURE_POPULATION_BASIS = "Please select the measure's Population basis prior to packaging.";
     private static final String MEASURE_DESCRIPTION_UNDERSCORE = "Measure name must not contain '_' (underscore).";
     public static final String MEASURE_TRANSFER_EMPTY_PACKAGE = "Unable to transfer selected measure. Issues were encountered with its package. Please re-package the measure before requesting a transfer to MADiE. If problem continues, please contact the help desk.";
-    public static final String MEASURE_TRANSFER_SUCCESS_MESSAGE = "Measure is being processed and transferred to MADiE.";
+    public static final String MEASURE_TRANSFER_SUCCESS_MESSAGE = "Measure is being processed and transferred to MADiE. You will receive an email with the transfer status.";
     public static final String MEASURE_TRANSFER_GENERIC_ERROR_MESSAGE = "Unable to transfer selected measure to MADiE, try again. If problem continues, please contact help desk.";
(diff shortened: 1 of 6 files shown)


src/main/java/mat/client/shared/MessageDelegate.java
```diff
@@ -211,7 +211,7 @@ public class MessageDelegate {
     private static final String MEASURE_POPULATION_BASIS = "Please select the measure's Population basis prior to packaging.";
     private static final String MEASURE_DESCRIPTION_UNDERSCORE = "Measure name must not contain '_' (underscore).";
     public static final String MEASURE_TRANSFER_EMPTY_PACKAGE = "Unable to transfer selected measure. Issues were encountered with its package. Please re-package the measure before requesting a transfer to MADiE. If problem continues, please contact the help desk.";
-    public static final String MEASURE_TRANSFER_SUCCESS_MESSAGE = "Measure is being processed and transferred to MADiE.";
+    public static final String MEASURE_TRANSFER_SUCCESS_MESSAGE = "Measure is being processed and transferred to MADiE. You will receive an email with the transfer status.";
```
RohitKandimalla, line 214: There are many reasons to believe that our automated generated emails may end up in Spam folder, idk if we want to address it here. to let users check their spam folder just in case.

RohitKandimalla (approved):

src/main/java/mat/client/util/FeatureFlagConstant.java
```diff
@@ -6,4 +6,5 @@ public interface FeatureFlagConstant {
     String FHIR_BONNIE = "FhirBonnie";
     String MADIE = "MADIE";
     String MADIE_QDM = "MADIE_QDM";
+    String MADIE_QDM_RE_TRANSFER = "MADIE_QDM_RE_TRANSFER";
```
sb-prateekkeerthi, line 9: In the ticket it is mentioned that to use the feature flag name as: enableQdmRepeatTransfer, shouldn't this be the name for the feature flag

src/main/java/mat/client/shared/MessageDelegate.java
adongare: that is a good point. I'll talk to Nicole.

adongare: okay, we'll be adding the note in the user guide about it which hardly people read.

